PR #8, opened 2020-04-02
Use bash so substitutions work for "nmake"

rkeene:
```
root@ac06aa16deab:/opt/msvc/bin/x86# /bin/sh /opt/msvc/bin/x86/nmake
/opt/msvc/bin/x86/nmake: 3: /opt/msvc/bin/x86/nmake: Bad substitution
root@ac06aa16deab:/opt/msvc/bin/x86# /bin/bash /opt/msvc/bin/x86/nmake

Microsoft (R) Program Maintenance Utility Version 14.25.28612.0

NMAKE : fatal error U1064: MAKEFILE not found and no target specified
Stop.
root@ac06aa16deab:/opt/msvc/bin/x86# 
```

### DIFF
--- a/wrappers/nmake
+++ b/wrappers/nmake
@@ -1,3 +1,3 @@
-#!/bin/sh
+#!/bin/bash
 . $(dirname $0)/msvcenv.sh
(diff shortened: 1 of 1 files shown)


mstorsjo:
Oh, oops - thanks for fixing!

Issue #10, opened 2020-04-15
WINEPATH does not include necessary directories


Cryspia:
In the tool wrappers, there are two sources of **WINEPATH** enviroment variables, but both may cause problems:

1. For _nmake_, the **WINEPATH** comes from **BINDIR**.
To my understanding, **SDKBINDIR** should also be included. Otherwise, _nmake_ cannot correctly find _rc_. You can try to compile [putty](https://github.com/github/putty) to regenerate an error that complains about missing _rc_.

2. For other wrappers, the **WINEPATH** is exported in _msvcenv.sh_ and is set to **$MSVCDIR\\bin\\Hostx64\\x64**
This directory is fixed to **Hostx64\\x64** regardless of which arch we are targeting. To the best of my knowledge, the folder should depend on the target arch, and it is better to use original directories (**SDKBINDIR** and **BINDIR**) instead of the wrapper directories in case that some other exes and dlls are required.

mstorsjo:
> In the tool wrappers, there are two sources of **WINEPATH** enviroment variables, but both may cause problems:
> 
> 1. For _nmake_, the **WINEPATH** comes from **BINDIR**.
>    To my understanding, **SDKBINDIR** should also be included. Otherwise, _nmake_ cannot correctly find _rc_. You can try to compile [putty](https://github.com/github/putty) to regenerate an error that complains about missing _rc_.

Yes, this is true

> 2. For other wrappers, the **WINEPATH** is exported in _msvcenv.sh_ and is set to **$MSVCDIR\bin\Hostx64\x64**
>    This directory is fixed to **Hostx64\x64** regardless of which arch we are targeting. To the best of my knowledge, the folder should depend on the target arch, and it is better to use original directories (**SDKBINDIR** and **BINDIR**) instead of the wrapper directories in case that some other exes and dlls are required.

No, this part is actually intentionally the way it is.

For older MSVC versions (2017), the cross compiler dirs (e.g. Hostx64/arm64, or anything other than x64) is missing a bunch of DLLs, e.g. mspdbcore.dll and msvcdis140.dll. So whenever running tools from Hostx64/whatever, Hostx64/x64 is expected to be in the path. Same thing if Hostx86/whatever is used, then Hostx86/x86 would have to be available in the path, for those DLLs to be found.

But I do see that this does conflict a bit with the nmake case. I'd suggest keeping the Hostx64/x64 case last in the path, but adding SDKBINDIR and BINDIR in front of it probably is good.

[...]

PR #11, opened 2020-04-15
Fix WINEPATH to include SDK bin directory

Cryspia:
The WINEPATH not correct include the vc bin directory and the SDK bin
directory, so that the compiler executables (for example nmake) can
correctly find all other tools. #10


mstorsjo:
Thanks! I squashed the commits manually (to retain proper authorship as github's squash+merge changes it), removed stray(?) changes to file permissions (at least the nmake one needs +x) and adjusted the commit message.

PR #18, opened 2020-08-09
Parallelize downloads of packages and add socket timeout.

oprypin:
At face value the parallelization makes no sense, but I've been seeing horrible download performance and stalling, so hopefully at least one of the N files will be properly downloading at a given time.

The socket timeout is needed to weed out stalled packages. I am consistently getting permanent stalling in downloads. If nothing happened on the socket in 15 seconds, it probably never will.

### DIFF
--- a/vsdownload.py
+++ b/vsdownload.py
@@ -18,9 +18,11 @@ import argparse
 import functools
 import hashlib
 import os
+import multiprocessing.pool
 import simplejson
 import six
 import shutil
+import socket
 import subprocess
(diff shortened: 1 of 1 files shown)


mstorsjo:
Thanks - the code for this does look good, but just to make sure - have you tested it with both python 2 and 3?

oprypin:
Yes, I checked that it works in both.

mstorsjo:
Thanks!

PR #25, opened 2021-06-14
Broken when source files contain spaces

marciosmiderle:
Hi, thank you for the project :-)

I know source files should have no spaces, but there is some projects that do this way. So, it is necessary to set IFS to "" in order to not break a file argument into two.

### DIFF
--- a/wrappers/wine-msvc.sh
+++ b/wrappers/wine-msvc.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+IFS=""
 EXE=$1
 shift
 ARGS=()


mstorsjo:
Hi, thanks for your pull request. Can you give a concete minimal standalone example of a case that currently doesn't break? I tested creating a test C file named `hello world.c` and it seems to work just fine:

```
$ ~/msvc2019-16.11-preview1/bin/x86/cl -c "hello world.c"
Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30129.3 for x86
 
hello world.c
```

marciosmiderle:
Inside a Makefile:

`
hello:
	~/Programas/msvc-wine/v142/bin/x86/cl  /nologo /TP /Fohello_world.cpp.obj /Fddir.dir/ /FS -c "/tmp/teste/hello world.cpp"
`

Occurs only when the filename is absolute, and it is the case when generating Makefiles with cmake.

mstorsjo:
Thanks, this I'm able to reproduce.

I'd rather fix this differently though, instead of changing `IFS` I'd fix the missing case of quoting, I pushed a different fix - see 84a46610378ef6118e60fc5d53a161f12e8d1b76. But thanks for your contribution!

Issue #29, opened 2021-10-11
Backslashes confuse CMake's dependency scanning


maharmstone:
CMake's dependency scanning for MSVC involves it passing the `/showIncludes` flag to `cl`, which results in it printing 

```
Note: including file: \opt\msvc\vc\tools\msvc\14.29.30133\include\string
Note: including file:  \opt\msvc\vc\tools\msvc\14.29.30133\include\yvals_core.h
Note: including file:   \opt\msvc\vc\tools\msvc\14.29.30133\include\vcruntime.h
```

etc. to stdout. Unfortunately the backslashes confuse it, meaning that it ends up recompiling the whole project every time.

Changing the third line of `/opt/msvc/bin/x64/cl` to

```
$(dirname $0)/wine-msvc.sh $BINDIR/cl.exe "$@" | sed "s/\\\\/\//g" -
```

fixes this, but it's not a very elegant way of doing it. It probably should say "for every line beginning 'Note: including file:', replace backslashes with slashes" - however you express that in sed.

mstorsjo:
Have a look at (the topmost commit in) https://github.com/mstorsjo/msvc-wine/commits/cmake - that contains an example of building a lightly patched cmake and ninja, which then should work fine with respect to dependency tracking. Unfortunately, those patches aren't entirely fit to upstream...

mstorsjo:
Thanks! I actually have a similar patch in a local test branch since some time (with a few minor tweaks to the regex, mine is `'/^Note:/s,\\,/,g'` (but I had applied it in `wine-msvc.sh`, while keeping it in the `cl` wrapper only is a better idea.

I'll try to refresh my memory with what I concluded about it and why I didn't go forward with it, whether it interfered with some of my other build setups or not... But meanwhile, I don't mind taking a patch for it!

This issue was discussed a bit in https://github.com/ninja-build/ninja/issues/1735, and in https://github.com/ninja-build/ninja/issues/1735#issuecomment-604089405 I seem to have concluded that it seemed to work fine that far...

darksylinc:
Hi! First: FANTASTIC WORK! It's amazing.

Second: Happy new year.

Third: I'm hit by this issue and the proposed solution works (i.e. ninja no longer starts from scratch) but I'm seeing a similar problem with RC. The generated rules.ninja contain:

```ninja
#############################################
# Rule for compiling RC files.

rule RC_COMPILER__OgreMain_Release
  depfile = $DEP_FILE
[...]

PR #33, opened 2022-01-01
Add native build instructions and FAQ

darksylinc:
The project is amazing but it is incredibly undocumented.

I hope this PR gets accepted to bring some lights to those seeking how to set it up, specially outside of docker.

Cheers and happy new year

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,3 +29,107 @@ The installer scripts also work fine without docker; just run the following two 
 
 The unpacking requires recent versions of msitools (0.98) and libgcab
 (1.2); sufficiently new versions are available in e.g. Ubuntu 19.04.
+
+
+# Build instructions for local installation
+
+The following instructions are for setting up MSVC without docker.
+
+## Prerequisites
+
+```bash
+apt-get update
+apt-get install -y wine64-development python msitools python-simplejson python-six ca-certificates winbind
+```
+
+## Installation
+
+We're going to install it into `~/my_msvc` to avoid needing root privileges on a non-contained system.
+
+```bash
+# Download and unpack MSVC
+./vsdownload.py --dest ~/my_msvc/opt/msvc
+# Add wrapper scripts, do minor cleanup of the unpacked MSVC installation
+./install.sh ~/my_msvc/opt/msvc
+
+# Custom CMake
+git clone https://gitlab.kitware.com/mstorsjo/cmake
+cd cmake
+git checkout 8c1957b8026c60d50a705555a3a212c9217c0a2e
+mkdir build 
+cd build
+../configure --prefix=~/my_msvc/opt/cmake --parallel=$(nproc) -- -DCMAKE_USE_OPENSSL=OFF
+make -j$(nproc)
+make install
+
+# Run wine at least once
+wineserver -k # kills server (optional)
+wineserver -p
+wine64 wineboot
+```
+
+> **_Note:_** the installation path will be hardcoded in the installed `msvcenv.sh`, thus moving the folder to another location is not recommended.
+
+### Setting up your project with CMake
+
+You need to set the path to prioritize our custom CMake, and also to see our MSVC installation.
+After that we just run CMake command with a few extra settings:
+
+```bash
+export PATH=~/my_msvc/opt/cmake/bin:$PATH
+export PATH=~/my_msvc/opt/msvc/bin/x64:$PATH
+CC=cl CXX=cl cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_PROGRAMS=ON -DCMAKE_SYSTEM_NAME=Windows -DCMAKE_CROSSCOMPILING=ON
+make
+```
+
+# FAQ
+
+## Does it run on Ubuntu 18.04 LTS?
+
+Yes, but the install scripts won't work because `msitools` is too old. You'll need to install either via Docker or on a real Ubuntu 20.04 LTS machine; and later copy paste the files under `/opt/msvc`.
+
+## Does it work with CMake?
+
+Yes, but a custom version is needed or else CMake will complain that `/opt/msvc/bin/x64/cl` can't build a simple program.
+
+It also fixes the `Ninja` Generator which otherwise has trouble finding RC.
+
+## Can it build Debug versions?
+
+Yes, but there may be troubles. From wine errors appearing in the logs/console to problems when trying to launch it because of missing debug redistributables.
+You will also have to install winbind (see next item).
+
+For the best out-of-the-box experience build in Release mode.
+
+## fatal error C1902: Program database manager mismatch; please check your installation
+
+You need winbind: `sudo apt install winbind`
+
+Issue is [being tracked](https://github.com/mstorsjo/msvc-wine/issues/6).
+
+## What generators work with CMake?
+
+The following generators were tested and known to work:
+
+ - Ninja
+ - Unix Makefiles
+
+Other generators are untested and may or may not work. Use it at your own peril.
+
+## Do I _need_ CMake to use msvc-wine?
+
+No. Using Ninja or GNU Make directly should work.
+
+## I get `ninja: error: build.ninja:225: bad $-escape (literal $ must be written as $$)`
+
+Visual Studio can switch between Debug/Release/RelWithDebInfo/etc at build time in the IDE.
+
+It's slightly common for CMake projects to use `$(CONFIGURATION)` macro from Visual Studio to resolve commands to each intended configuration automatically.
+
+However generators like `Ninja`/`Unix Makefiles` can only target one configuration at a time.
+
+You'll have to edit the CMake script to use `${CMAKE_BUILD_TYPE}` instead when using `Ninja`/`Unix Makefiles` generators (which is extremely rare to use in an actual Windows environment).
+
+Note the script may have other hardcoded commands which use `$(...)` syntax that make no sense when using generators other than Visual Studio; and will need to be fixed accordingly.
+
+This is not an msvc-wine bug.


INSTRUCTIONS.native.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Does it work with CMake?
+
+Yes, but a custom version is needed or else CMake will complain that `/opt/msvc/bin/x64/cl` can't build a simple program.
+
+## Can it build Debug versions?
+
+Yes, but there may be troubles. From wine errors appearing in the logs/console to problems when trying to launch it because of missing debug redistributables.
+
+For the best experience build in Release mode
```
mstorsjo, line 17: Sentence missing a trailing period.

INSTRUCTIONS.native.md
```diff
@@ -0,0 +1,92 @@
...
+
+## I get an "access denied" error when building RC via CMake.
+
+`Ninja` Generator has a problem when setting up RC. Use `Unix Makefile`s instead
```
mstorsjo, line 21: Missing trailing period. Also overall I didn't even think unix makefiles work with MSVC - that's surprising/news to me.

INSTRUCTIONS.native.md
```diff
@@ -0,0 +1,92 @@
...
+
+There is a [custom version of Ninja](https://github.com/mstorsjo/ninja/commits/cross-msvc) to deal with MSVC cross compiling but I haven't checked it to see if it fixes the problem.
```
mstorsjo, line 23: I'd kinda prefer to at least look into the issue before adding a FAQ entry saying "I haven't checked"...

(I haven't had time to look into your other comment on this yet - I'm a bit surprised that I haven't run into it, but maybe the cmake projects I've built haven't included any RC files?)

INSTRUCTIONS.native.md
```diff
@@ -0,0 +1,92 @@
...
+
+# FAQ
+
+## Does it run on Ubuntu 18.04 LTS?
+
+Yes, but the install scripts won't work because `msitools` is too old. You'll need to install either via Docker or on a real Ubuntu 20.04 LTS machine; and later copy paste the files under `/opt/msvc`
+
+## Does it work with CMake?
+
+Yes, but a custom version is needed or else CMake will complain that `/opt/msvc/bin/x64/cl` can't build a simple program.
```
mstorsjo, line 11: IIRC there's an issue with discussion about this as well, with indications that it might work kinda fine with stock CMake too - it might be good to link to that too.

INSTRUCTIONS.native.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Installation
+
+We're going to install it into `~/my_msvc` to avoid needing root privileges on a non-contained system.
+You can change the location to wherever you want, even real `/opt/` if you want to deal with chown'ing all files to root before moving
+
+```bash
+# Install MSVC
+./vsdownload.py --dest ~/my_msvc/opt/msvc
+./install.sh ~/my_msvc/opt/msvc
```
mstorsjo, line 65: Here I think it'd be nice to have a separate comment for the `install.sh` line, singling them out more, e.g. like this:
```
# Download and unpack MSVC
./vsdownload.py ...
# Add wrapper scripts, do minor cleanup of the unpacked MSVC installation
./install.sh ...
```

README.md
```diff
@@ -29,3 +29,10 @@ The installer scripts also work fine without docker; just run the following two
 
 The unpacking requires recent versions of msitools (0.98) and libgcab
 (1.2); sufficiently new versions are available in e.g. Ubuntu 19.04.
+
+
+# Build instructions for native environments
+
+See [INSTRUCTIONS.native.md](INSTRUCTIONS.native).
```
mstorsjo, line 36: I'd prefer to have all of this included in the main README.md, instead of splitting out a separate file. I agree that the docker setup isn't what most people might be using in the end, but it's great as a sealed, self contained and reproducible example for how it works. So the main README could be 1. minimal docker example, 2. normal local installation examples 3. FAQ.

Also as name for this kind of installation, I'd prefer "local installation" or something else to denote the non-docker installation setup, rather than "for native environments".

mstorsjo (commented):
Thanks for the PR! It looks useful overall, but I'd like it a bit reshuffled and restructured before merging...

INSTRUCTIONS.native.md
darksylinc: I noticed the comment about vanilla CMake working but I tried it on Cmake version 3.22.1 and it was still broken.

I didn't try but I suspect that if I do the first CMake with the modified version (so that CMake compiles the test app) afterwards any later configure will work with any cmake version (this is a theory, I didn't try it yet)

INSTRUCTIONS.native.md
darksylinc: Done!

README.md
darksylinc: I wasn't exactly sure how to handle it.

The way I see there are two very distinct users who want very different things:

1. CI devs who want a container that "just works" to test MSVC projects still work on Windows without having to launch a Windows server.
   - A slight variation of those who want a CI for do stuff like godbolt (i.e. online compiler)
2. Devs who want to compile in a native environment. Often for iterating (i.e. build -> fix errors -> build -> launch -> fix errors -> build); which means tinkering with the compiler until it works with their IDE (btw it works with QtCreator! Although Code Autocomplete & Syntax highlighting breaks because Clang plugin doesn't notice it's an MSVC installation. I alternate between mingw64 for syntax highlighting, then to MSVC for building. It's good enough for me for now)

INSTRUCTIONS.native.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Does it work with CMake?
+
+Yes, but a custom version is needed or else CMake will complain that `/opt/msvc/bin/x64/cl` can't build a simple program.
+
+## Can it build Debug versions?
+
+Yes, but there may be troubles. From wine errors appearing in the logs/console to problems when trying to launch it because of missing debug redistributables.
+
+For the best experience build in Release mode.
```
mstorsjo, line 17: I think you should refer to info about PDB files and winbind below too, to inform about how to potentially make debug mode things work. (But the gist of it; stick to release mode if possible, is good to get across.)

INSTRUCTIONS.native.md
mstorsjo: I think we can remove this section about the RC issues now, by pointing the "custom cmake" at the newer commit that includes the cherrypicked fix for it.

INSTRUCTIONS.native.md
```diff
@@ -0,0 +1,93 @@
...
+## I get an "access denied" error when building RC via CMake.
+
+`Ninja` Generator has a problem when setting up RC. Use `Unix Makefile`s instead.
+
+There is a [custom version of Ninja](https://github.com/mstorsjo/ninja/commits/cross-msvc) to deal with MSVC cross compiling but I haven't checked it to see if it fixes the problem.
+
+## Project gets rebuild from scratch every time
+
+This is a [known issue](https://github.com/mstorsjo/msvc-wine/issues/29)
+Apply this patch:
```
mstorsjo, line 28: I'll try to revisit this issue now and see if I can merge a fix, so we don't need to suggest applying patches in the docs...

README.md
mstorsjo: Sure, but even if there are these two cases, I'd want this info to be available in the main readme, not hidden away in a separate document. So please merge it into the main readme, with the ordering I suggested (first simple docker example, then regular local install, then FAQ).

darksylinc:
OK I applied your suggestions.

Everything is merged. Docker, then Native, then FAQ.


I also the install instructions to point to your latest CMake fix.

I also updated the FAQ as per your feedback.

darksylinc:
Added another common error to the FAQ users may encounter

README.md
```diff
@@ -29,3 +29,117 @@ The installer scripts also work fine without docker; just run the following two
...
+
+```bash
+apt-get update
+apt-get install -y wine64-development python msitools python-simplejson python-six ca-certificates winbind
+```
+
+## Installation
+
+We're going to install it into `~/my_msvc` to avoid needing root privileges on a non-contained system.
+You can change the location to wherever you want, even real `/opt/` if you want to deal with chown'ing all files to root before moving
```
mstorsjo, line 48: Please end sentences with a period.

Also currently, the installation path is hardcoded in the installed `msvcenv.sh` so one can't easily e.g. install into a directory under the home directory and then move them with sudo. So I'd suggest maybe leaving out the last sentence altogether (or clarify that the installation directory currently isn't movable from the path it was installed into).

README.md
```diff
@@ -29,3 +29,117 @@ The installer scripts also work fine without docker; just run the following two
...
+
+```bash
+# Download and unpack MSVC
+./vsdownload.py --dest ~/my_msvc/opt/msvc
+# Add wrapper scripts, do minor cleanup of the unpacked MSVC installation
+./install.sh ~/my_msvc/opt/msvc
+
+# Custom CMake
+git clone https://gitlab.kitware.com/mstorsjo/cmake
+git checkout 8c1957b8026c60d50a705555a3a212c9217c0a2e
```
mstorsjo, line 58: This is missing a `cd cmake` step.

README.md
```diff
@@ -29,3 +29,117 @@ The installer scripts also work fine without docker; just run the following two
...
+
+# Run wine at least once
+wineserver -k # kills server (optional)
+wineserver -p
+wine64 wineboot
+```
+
+### Setting up your project with CMake
+
+You need to set the path to prioritize our custom CMake, and also to see our MSVC installation
```
mstorsjo, line 73: Sentence missing a period.

README.md
```diff
@@ -29,3 +29,117 @@ The installer scripts also work fine without docker; just run the following two
...
+
+# FAQ
+
+## Does it run on Ubuntu 18.04 LTS?
+
+Yes, but the install scripts won't work because `msitools` is too old. You'll need to install either via Docker or on a real Ubuntu 20.04 LTS machine; and later copy paste the files under `/opt/msvc`
+
+## Does it work with CMake?
+
+Yes, but a custom version is needed or else CMake will complain that `/opt/msvc/bin/x64/cl` can't build a simple program.
```
mstorsjo, line 91: The exact error seems to vary, sometimes it hangs at the "Detecting C compiler ABI info" stage, sometimes it errors out cleanly. It might be good to mention both symptoms.

README.md
```diff
@@ -29,3 +29,117 @@ The installer scripts also work fine without docker; just run the following two
...
+
+## Does `Ninja` work?
+
+Yes.
+
+##  Does `Unix Makefile` work?
+
+Yes.
+
+## Project gets rebuild from scratch every time
```
mstorsjo, line 116: This issue can skipped from the list, as issue #29 should be fixed now.

README.md
```diff
@@ -29,3 +29,117 @@ The installer scripts also work fine without docker; just run the following two
...
++$(dirname $0)/wine-msvc.sh $BINDIR/cl.exe "$@" | sed "/Note: including file:/ s/\\\\/\//g" -
+```
+
+## I get `ninja: error: build.ninja:225: bad $-escape (literal $ must be written as $$)`
+
+Visual Studio can switch between Debug/Release/RelWithDebInfo/etc at build time in the IDE.
+
+It's slightly common for CMake projects to use `$(CONFIGURATION)` macro from Visual Studio to resolve commands to each intended configuration automatically.
+
+However generators like `Ninja`/`Unix Makefile` can only target one configuration at a time.
```
mstorsjo, line 139: There's a `Ninja Multi-Config` generator available too, that should handle that setup.

README.md
```diff
@@ -29,3 +29,117 @@ The installer scripts also work fine without docker; just run the following two
...
+
+You'll have to edit the CMake script to use `${CMAKE_BUILD_TYPE}` instead when using `Ninja`/`Unix Makefile` generators (which is extremely rare to use in an actual Windows environment).
```
mstorsjo, line 141: I'd omit the parenthesis here, the judgement whether it's common or rare doesn't really add value here.

README.md
```diff
@@ -29,3 +29,117 @@ The installer scripts also work fine without docker; just run the following two
...
+
+You need winbind: `sudo apt install winbind`
+
+Issue is [being tracked](https://github.com/mstorsjo/msvc-wine/issues/6).
+
+## Does `Ninja` work?
+
+Yes.
+
+##  Does `Unix Makefile` work?
```
mstorsjo, line 112: Hmm, when first reading these two points (does ninja work, does unix makefiles work), my first reaction is that these are quite cmake specific - i.e. yes, the cmake ninja and unix makefiles generators do work, and gnu make and ninja handle it. But it's of course a general point too (ninja works fine with meson too, and gnu make works fine with e.g. ffmpeg).

So I'm undecided if we should make the points clearly more cmake specific ("Does `Ninja` work in cmake?") or more generic ("Does ninja work with these tools? - Yes, it's known to work with the modified cmake and unmodified meson. - Do unix makefiles work? - Yes, such a setup is known to work with the modfiied cmake and with ffmpeg's build system.")

Btw if you literally quote `Unix Makefiles`, note that the generator name ends with an `s`.

README.md
```diff
@@ -29,3 +29,117 @@ The installer scripts also work fine without docker; just run the following two
 
 The unpacking requires recent versions of msitools (0.98) and libgcab
 (1.2); sufficiently new versions are available in e.g. Ubuntu 19.04.
+
+
+# Build instructions for native environments
```
mstorsjo, line 34: I'd still prefer to use a different term than "native environments" as it's a quite unclear term in this case.

"Build instructions for local installation" maybe?

darksylinc: Agreed. I didn't realize "native" here can mean Windows or Linux non-Docker.

It can cause confusion. I went with your suggestion

README.md
darksylinc: Good points. I reworded these points.

README.md
darksylinc: I disagree here.

In my experience a small percentage of users are dumb and don't RTFM and may insist it's an msvc-wine bug. They need to be repeated explanations and stuff multiple times.

Sadly this also applies to developer-oriented tools like this project (where you'd think everyone is supposed to have critical thinking).

Maybe one day msbuild.exe can be emulated; but aside from being a big endeavour and out of scope for this project (is it?) CMake for Linux would also have to be build with msbuild.exe support; which is currently not (or maybe we should use CMake.exe instead via wine).

README.md
darksylinc: I didn't want to mention `Ninja Multi-Config` because I didn't test it. I don't know if it actually handles it correctly.

README.md
darksylinc: Done

README.md
darksylinc: The hang is still there even with the custom CMake. But it's rare.

It sounds like that's a different bug.

README.md
darksylinc: Oh! I didn't know! I reworded it to point this out.

mstorsjo:
Ok, I guess this looks fine now. I'd merge it with "squash and merge", but I guess the commit would look nicer if you'd rebase and squash it manually.

mstorsjo:
Thanks for your contribution!

PR #39, opened 2022-01-26
add double-quotes to shell scripts

panzi:
I think this should make it possible to install it to a path that contains spaces. (Haven't tested it, had no time for that today.)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -3,25 +3,25 @@
 set -e
 
 if [ $# -lt 1 ]; then
-    echo $0 {vc.zip sdk.zip target\|target}
+    echo "$0 {vc.zip sdk.zip target|target}"
     exit 0
 fi
 
 if [ $# -eq 3 ]; then
-    VC_ZIP=$(cd $(dirname $1) && pwd)/$(basename $1)
-    SDK_ZIP=$(cd $(dirname $2) && pwd)/$(basename $2)
+    VC_ZIP=$(cd "$(dirname "$1")" && pwd)/$(basename "$1")
+    SDK_ZIP=$(cd "$(dirname "$2")" && pwd)/$(basename "$2")
     DEST=$3
 else
     DEST=$1
 fi
-ORIG=$(cd $(dirname $0) && pwd)
+ORIG=$(cd "$(dirname "$0")" && pwd)
 
-mkdir -p $DEST
-cd $DEST
+mkdir -p "$DEST"
+cd "$DEST"
 DEST=$(pwd)
 
 if [ -n "$VC_ZIP" ]; then
-    unzip $VC_ZIP
+    unzip "$VC_ZIP"
 fi
 mv VC vc
 mv vc/Tools vc/tools
@@ -32,11 +32,11 @@ mv vc/tools/MSVC vc/tools/msvc
 # /DEFAULTLIB:"OLDNAMES", which lld-link doesn't find on a case sensitive
 # filesystem. Therefore add matching case symlinks for this, to allow
 # linking MSVC built objects with lld-link.
-cd $(echo vc/tools/msvc/* | awk '{print $1}')/lib
+cd "$(echo vc/tools/msvc/* | awk '{print $1}')/lib"
 for arch in x86 x64 arm arm64; do
-    cd $arch
+    cd "$arch"
     for i in libcmt libcmtd msvcrt msvcrtd oldnames; do
-        ln -s $i.lib $(echo $i | tr [a-z] [A-Z]).lib
+        ln -s "$i.lib" "$(echo "$i" | tr '[a-z]' '[A-Z]').lib"
     done
     cd ..
 done
@@ -44,9 +44,7 @@ cd ../bin
 # vctip.exe is known to cause problems at some times; just remove it.
 # See https://bugs.chromium.org/p/chromium/issues/detail?id=735226 and
 # https://github.com/mstorsjo/msvc-wine/issues/23 for references.
-for i in $(find . -iname vctip.exe); do
-    rm $i
-done
+find . -iname vctip.exe -exec rm {} \;
 cd "$DEST"
 
 if [ -d kits/10 ]; then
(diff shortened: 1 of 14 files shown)


mstorsjo:
> I think this should make it possible to install it to a path that contains spaces. (Haven't tested it, had no time for that today.)

Please do test, I don't want to merge this blindly if it wouldn't actually fix the thing it sets out to fix (even if more quoting generally is good).

panzi:
Ok, tested it now. There was indeed some place where the spaces weren't properly handled yet. I've added `replacevars.py` to do correct replacements in `msvcenv.sh`. This could probably also be done with perl or awk, but I know how to do it in Python.

install.sh
```diff
@@ -3,25 +3,25 @@
 set -e
 
 if [ $# -lt 1 ]; then
-    echo $0 {vc.zip sdk.zip target\|target}
+    echo "$0 {vc.zip sdk.zip target|target}"
```
mstorsjo, line 6: This particular line change is made mostly for consistency right - spaces in `$0` shouldn't be an issue here?

wrappers/wine-msvc.sh
```diff
@@ -4,7 +4,7 @@ shift
 ARGS=()
 while [ $# -gt 0 ]; do
 	a=$1
-	case $a in
+	case "$a" in
```
mstorsjo, line 7: Hmm, I guess this particular case here is something that is done just in case, but I don't see where a path with spaces would be passed as argument here?

install.sh
```diff
@@ -32,50 +32,52 @@ mv vc/tools/MSVC vc/tools/msvc
 # /DEFAULTLIB:"OLDNAMES", which lld-link doesn't find on a case sensitive
 # filesystem. Therefore add matching case symlinks for this, to allow
 # linking MSVC built objects with lld-link.
-cd $(echo vc/tools/msvc/* | awk '{print $1}')/lib
+cd "$(echo vc/tools/msvc/* | awk '{print $1}')/lib"
 for arch in x86 x64 arm arm64; do
-    cd $arch
+    cd "$arch"
```
mstorsjo, line 37: The change on this line is kinda superfluous IMO (I know you did it for consistency though), as `$arch` only is either of the 4 single words listed above.

install.sh
```diff
@@ -32,50 +32,52 @@ mv vc/tools/MSVC vc/tools/msvc
...
     for i in libcmt libcmtd msvcrt msvcrtd oldnames; do
-        ln -s $i.lib $(echo $i | tr [a-z] [A-Z]).lib
+        ln -s "$i.lib" "$(echo "$i" | tr '[a-z]' '[A-Z]').lib"
```
mstorsjo, line 39: Also unnecessary

mstorsjo:
> I've added `replacevars.py` to do correct replacements in `msvcenv.sh`. This could probably also be done with perl or awk, but I know how to do it in Python.

Hmm, I'm not a huge fan of adding an extra script just for doing those replacements. But in any case, I was planning another refactoring, to not hardcode `BASE` and `BASE_UNIX` as absolute paths, but calculate them at runtime, to allow moving around the installed toolchain. With that change in place (I just pushed it as f983e0be893f588a3360e7384dde930ea23b6e2b), there shouldn't be any need to change how the replacements are done, and the preexisting `sed` based replacement should work just fine.

> I've squashed the fixes to my earlier commit so its still only one commit so I don't pollute your history.

Thanks, that's appreciated, that's my preferred form of working too!

install.sh
panzi: Yes, that one I made for consistency.

Unimportant tangent: Although if you want to be 100% correct: `echo` *does* accept one flag `-n`. At least GNU echo does. This flag tells echo to not print a newline. So if someone would exec `install.sh` with the name `-n` (this is possible, see `man execve`) it would create unexpected output. But at that point someone is trying already hard just to remove the newline. Why would anyone do that? A solution would be `printf '%s {vc.zip sdk.zip target|target}' "$0"` (but who cares about that).

wrappers/wine-msvc.sh
panzi: Yes, at this point I was just being consistent. Better to have unnecessary quotes than forgetting them somewhere.

install.sh
panzi: True. Again my thinking was better to have unnecessary quotes than forgetting them somewhere. :smile:

install.sh
panzi: True. Thing is my editor's linter even complains about unquoted strings anywhere, so I just mechanically add them. :smile:

panzi:
Ah, I see there are new (conflicting) changes. I'll rebase my branch to that.

mstorsjo:
> > I've added `replacevars.py` to do correct replacements in `msvcenv.sh`. This could probably also be done with perl or awk, but I know how to do it in Python.
> 
> Hmm, I'm not a huge fan of adding an extra script just for doing those replacements. [...] With that change in place (I just pushed it as [f983e0b](https://github.com/mstorsjo/msvc-wine/commit/f983e0be893f588a3360e7384dde930ea23b6e2b)), there shouldn't be any need to change how the replacements are done, and the preexisting `sed` based replacement should work just fine.

I'll reiterate this point just to make it clear. I don't want to add a separate script to do the replacements, when the current in-line sed command works just fine.

I understand that you are on a mission to add doublequotes everywhere to appease your personal coding standard. I'm open to adding them where they make a practical difference. I can tolerate adding them in superfluous places for consistency. But I don't want a separate standalone script for doing replacements where the current sed oneliners work just fine. (Yes I know you can come up with some theoretical issue with surprising characters in `MSVCVER` and `SDKVER`, but that's clearly out of scope here, and I'm not adding `replacevars.py` for taking care of that.)

panzi:
Oh, sorry, I somehow totally missed that comment and only saw the inline comments. Would you instead of the mixture of bash, Python, and Perl scripts perhaps prefer just a Python script to do the job of `install.sh`?

> I understand that you are on a mission to add doublequotes everywhere to appease your personal coding standard.

I'm not. I just mechanically did when fixing the issue. But I'm happy to adopt any coding standard of someone else's project when I'm aware of it.

mstorsjo:
> Oh, sorry, I somehow totally missed that comment and only saw the inline comments. Would you instead of the mixture of bash, Python, and Perl scripts perhaps prefer just a Python script to do the job of `install.sh`?

No don't feel the desire to rewrite these scripts in another language at the moment. I don't mind the existing setup at all (I hadn't even reflected on the fact that it does indeed contain a couple separate perl scripts). Just keep this PR as it is, but remove the new `replacevars.py` script and restore the original calls to `sed` (possibly with extra quoting of the regexes if you prefer that), and this should be fine with me.

mstorsjo:
> > Oh, sorry, I somehow totally missed that comment and only saw the inline comments. Would you instead of the mixture of bash, Python, and Perl scripts perhaps prefer just a Python script to do the job of `install.sh`?
> 
> No don't feel the desire to rewrite these scripts in another language at the moment. I don't mind the existing setup at all (I hadn't even reflected on the fact that it does indeed contain a couple separate perl scripts). Just keep this PR as it is, but remove the new `replacevars.py` script and restore the original calls to `sed` (possibly with extra quoting of the regexes if you prefer that), and this should be fine with me.

Ping - I'd be fine with the rest of this PR if you'd skip the rewrite/addition of a new script and just keep this to adding more quoting.

TheComputerGuy96:
@panzi Do you have any plans to rebase/update this MR? If not, then I'll make my own with an updated version of your commit

panzi:
@TheComputerGuy96 Yeah, I don't need this at the moment so I didn't maintain this patch.

TheComputerGuy96:
I just pushed a branch with an updated version of your commit: https://github.com/TheComputerGuy96/msvc-wine/tree/double-quotes (you can either cherry-pick it or I'll make another MR)

mstorsjo:
> I just pushed a branch with an updated version of your commit: https://github.com/TheComputerGuy96/msvc-wine/tree/double-quotes (you can either cherry-pick it or I'll make another MR)

Thanks, this is appreciated. If you'd open a new PR with this, it'll get a run through the CI, and I can merge it.

PR #40, opened 2022-02-14
ported install.sh to Python

panzi:
This also removes the dependency to Perl, reducing the number of different languages used.

Just in case you are interested in this. If not, just ignore this. :smile:  
Also: I've of course added you to the copyright header of the new file, since it is very much based on your work. Though my version of `lowercase` is much simpler.


mstorsjo:
> This also removes the dependency to Perl, reducing the number of different languages used.
> 
> Just in case you are interested in this. If not, just ignore this. 😄 

Yeah - no thanks - the existing mixture of languages isn't that much of an issue right now, IMO.

PR #43, opened 2022-03-10
install: Skip over non-installed architectures

real-or-random:
This makes it possible to download only specific workloads such as
Microsoft.VisualStudio.Workload.VCTools, which does not include arm
or arm64.

I tested this locally (and which some "echo"s to see if it selects
the right directories, and it works for me.


mstorsjo:
Thanks!

PR #44, opened 2022-03-14
Change "$0" to read the executed source location to "${BASH_SOURCE[0]}

falhumai96:
Change `$0` to read the executed source location to `${BASH_SOURCE[0]}`, as the former code would not work in the case of sourcing the script because `$0` will have the value of the sourcer script.

This Stack Overflow answer thoroughly explains the difference between `$0` and `${BASH_SOURCE[0]}`: https://stackoverflow.com/a/35006505, as well as the linked answer within the above answer: https://stackoverflow.com/a/29835459.

### DIFF
--- a/wrappers/msvcenv.sh
+++ b/wrappers/msvcenv.sh
@@ -2,7 +2,7 @@
 
 SDK=kits\\10
 SDK_UNIX=kits/10
-BASE_UNIX=$(cd "$(dirname "$0")"/.. && pwd)
+BASE_UNIX=$(cd "$(dirname "${BASH_SOURCE[0]}")"/.. && pwd)
 # Support having the wrappers in a directory one or two levels below the
 # installation directory.
 if [ ! -d "$BASE_UNIX/vc" ]; then


mstorsjo:
This looks reasonable, but I wonder when you ever would want to source the frontend scripts like e.g. `wrappers/cl` (other than msvcenv.sh - I totally get why you'd want to source that one) or `install.sh`.

Or is the original story that you ran into issues with `$0` in msvcenv.sh and realized that should be `${BASH_SOURCE[0]}`, and you then went ahead and changed all the other scripts too just for consistency?

falhumai96:
> This looks reasonable, but I wonder when you ever would want to source the frontend scripts like e.g. `wrappers/cl` (other than msvcenv.sh - I totally get why you'd want to source that one) or `install.sh`.
> 
> Or is the original story that you ran into issues with `$0` in msvcenv.sh and realized that should be `${BASH_SOURCE[0]}`, and you then went ahead and changed all the other scripts too just for consistency?

Yes to the consistency and where I ran into issues. I am building a project where I combine CCCL and MSVC-Wine projects, namely `cccl-msvc-wine`, and I ran into issues when trying to source `msvcenv.sh` because `$0` was changed. The reason why I wanted to source that script is that I wanted to implement a functionality to `print-search-dirs` in CCCL by processing `LIB` environment variables from MSVC-Wine as well as any library path from the `-L` argument.

However, this current commit is not passing as I am having a `bad substitution` bug on `install.sh`, as I forgot that this script is `sh` and not `bash`.

falhumai96:
> > This looks reasonable, but I wonder when you ever would want to source the frontend scripts like e.g. `wrappers/cl` (other than msvcenv.sh - I totally get why you'd want to source that one) or `install.sh`.
> > Or is the original story that you ran into issues with `$0` in msvcenv.sh and realized that should be `${BASH_SOURCE[0]}`, and you then went ahead and changed all the other scripts too just for consistency?
> 
> Yes to the consistency and where I ran into issues. I am building a project where I combine CCCL and MSVC-Wine projects, namely `cccl-msvc-wine`, and I ran into issues when trying to source `msvcenv.sh` because `$0` was changed. The reason why I wanted to source that script is that I wanted to implement a functionality to `print-search-dirs` in CCCL by processing `LIB` environment variable from MSVC-Wine as well as any library path from the `-L` argument.
> 
> However, this current commit is not passing as I am having a `bad substitution` bug on `install.sh`, as I forgot that this script is `sh` and not `bash`. Debugging this issue now.

You know what, as my requirement is only for `msvcenv.sh`, I'll revert the rest and keep `msvcenv.sh` change(s).

mstorsjo:
> You know what, as my requirement is only for `msvcenv.sh`, I'll revert the rest and keep `msvcenv.sh` change(s).

Sounds good, I was just about to suggest that.

mstorsjo:
Thanks!

PR #54, opened 2022-11-24
Changes to install to make it work with clang-cl /winsysroot by default

tru:
This symlinks some subdirs instead of renaming them and makes sure there is a "Windows Kits" directory.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -37,9 +37,10 @@ DEST=$(pwd)
 if [ -n "$VC_ZIP" ]; then
     unzip $VC_ZIP
 fi
-mv VC vc
-mv vc/Tools vc/tools
-mv vc/tools/MSVC vc/tools/msvc
+ln -s kits "Windows Kits"
+ln -s VC vc
+ln -s Tools vc/tools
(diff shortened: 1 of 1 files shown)


mstorsjo:
LGTM, thanks!

FWIW, I appreciate that we keep the lowercased intermediate pathnames too. The background for using lowercase there, instead of just setting `INCLUDE` and `LIB` to the correctly cased paths (in `msvcenv.sh`), is that older versions of MSVC had a bug, where the paths of some dependencies (as output by `/showincludes`) would be totally lowercased, even if the paths contained uppercase letters originally. Due to this, any native linux (case sensitive) build system would rebuild everything all the time (since the dependencies pointed at headers at paths that didn't exist). Lowercasing all the internal paths in the MSVC install avoided much of the impact of this bug - as long as the project itself doesn't use uppercase letters in paths/header names. This was fixed in MSVC 2019 16.0 in https://developercommunity.visualstudio.com/t/showIncludes-lowercases-some-path-segme/233871. (Which at this point is a while ago already - but it's still possible to install MSVC 2017 too.)

PR #55, opened 2022-11-25
github: Test using the Clang /winsysroot parameter for pointing out the SDK

mstorsjo:
@tru Does this seem reasonable - in particular the comment? (The test uses Clang/LLD 14, since that's readily available in the GHA ubuntu 22.04 images.)


tru:
Yep looks fine to me! Good to have it tested so that it doesn't break if the layout changes.

mstorsjo:
> Yep looks fine to me! Good to have it tested so that it doesn't break if the layout changes.

Absolutely! Having automated testing is always a relief, knowing that you don't need to think of testing that aspect manually again.

mstorsjo:
(Apparently github doesn't close PRs automatically when manually pushing any longer?)

PR #59, opened 2023-04-11
Allow to use vsdownload on Windows

glandium:
The script can actually be useful to download VS on Windows too. When doing so, use the Windows-native msiexec tool to extract the MSIs, instead of msiextract.

And because of how open files are handled on Windows, use a context manager for the subprocess.check_call output, so that if it throws an exception, the file is not left open.

### DIFF
--- a/vsdownload.py
+++ b/vsdownload.py
@@ -511,9 +511,12 @@ def unpackWin10SDK(src, payloads, dest):
         if name.endswith(".msi"):
             print("Extracting " + name)
             srcfile = os.path.join(src, name)
-            log = open(os.path.join(dest, "WinSDK-" + getPayloadName(payload) + "-listing.txt"), "w")
-            subprocess.check_call(["msiextract", "-C", dest, srcfile], stdout=log)
-            log.close()
+            if sys.platform == "win32":
+                cmd = ["msiexec", "/a", srcfile, "/qn", "TARGETDIR=" + os.path.abspath(dest)]
+            else:
+                cmd = ["msiextract", "-C", dest, srcfile]
+            with open(os.path.join(dest, "WinSDK-" + getPayloadName(payload) + "-listing.txt"), "w") as log:
+                subprocess.check_call(cmd, stdout=log)
 
 def extractPackages(selected, cache, dest):
     makedirs(dest)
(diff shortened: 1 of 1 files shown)


vsdownload.py
```diff
@@ -538,7 +541,10 @@ def moveVCSDK(unpack, dest):
     # files to be removed.
     makedirs(os.path.join(dest, "kits"))
     mergeTrees(os.path.join(unpack, "VC"), os.path.join(dest, "VC"))
-    mergeTrees(os.path.join(unpack, "Program Files", "Windows Kits", "10"), os.path.join(dest, "kits", "10"))
+    if sys.platform == "win32":
+        mergeTrees(os.path.join(unpack, "Windows Kits", "10"), os.path.join(dest, "kits", "10"))
+    else:
+        mergeTrees(os.path.join(unpack, "Program Files", "Windows Kits", "10"), os.path.join(dest, "kits", "10"))
```
mstorsjo, line 547: So there's a difference between the Windows native `msiexec` and the third party `msiextract`, where the latter includes an extra `Program Files/` toplevel directory here, while the former doesn't? That's tricky...

Can we reduce the amount of duplicated code here a bit, by doing e.g. this?
```
kitsPath = unpack
if sys.platform == "win32:
    kitsPath = os.path.join(kitsPath, "Program Files")
kitsPath = os.path.join(kitsPath, "Windows Kits", "10")
mergeTrees(kitsPath, os.path.join(dest, "kits", "10"))
```

mstorsjo (commented):
Thanks for the PR - that's a usecase I hadn't thought about! The patch looks mostly reasonable, just one  minor comment.

I presume the usecase is that this is faster and/or more customizable than using the build tools installer?

glandium:
> I presume the usecase is that this is faster and/or more customizable than using the build tools installer?

Yes, we're planning to use this to bootstrap the bits of VS required to build Firefox. We already use the script for cross compilation on CI, but we're going to use it for builds on developer machines, including Windows.

mstorsjo:
LGTM, thanks for your contribution! And it's nice to hear that you've found this useful enough for Firefox builds :-)

PR #60, opened 2023-04-13
Execute wine wrapped commands with WINEDEBUG=-all

mstorsjo:
This should silence all wine debug messages, instead of just filtering out some of them.

If building with MSVC with /showDependencies, the output from /showIncludes gets printed on stderr instead of stdout, and then wine debug messages may get intermixed with the /showIncludes output, breaking dependency handling.

CC @huangqinjin - does this work for you?

### DIFF
--- a/wrappers/wine-msvc.sh
+++ b/wrappers/wine-msvc.sh
@@ -38,5 +38,5 @@ while [ $# -gt 0 ]; do
 	ARGS+=("$a")
 	shift
 done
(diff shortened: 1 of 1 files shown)


mstorsjo:
> @mstorsjo Tested along with PR #61 , no more recompiling issued by Ninja. These PRs work perfectly on my project.

Thanks for testing!

> Small suggestion: `sed` supports multiple expressions separated by `;`. Can it increase through output slightly with less pipe?
> 
> And the `sed` expressions are exactly same for `stdout` and `stderr`, define a variable for the expressions could be less verbose.

Feel free to make a PR for such a change!

Issue #45, opened 2022-03-14
`mt.exe` issue with CMake


falhumai96:
Hello @mstorsjo,

I was trying to build a C/C++ project a while ago (not sure whether it was ZLib or LLVM, but pretty sure it was a CMake-based project), and CMake was failing on `mt.exe`. However, if `mt.exe` ran again after failing the first time, CMake would pass.

I was about to make the following change to the script, but was hesitant whether or not this is a hacky fix:

```
diff --git a/wrappers/mt b/wrappers/mt
index 30349ed..f6f359f 100755
--- a/wrappers/mt
+++ b/wrappers/mt
@@ -1,3 +1,7 @@
 #!/bin/bash
 . $(dirname $0)/msvcenv.sh
 $(dirname $0)/wine-msvc.sh $SDKBINDIR/mt.exe "$@"
+
+if [ $? -ne 0 ]; then
+    $(dirname $0)/wine-msvc.sh $SDKBINDIR/mt.exe "$@"
+fi
```

falhumai96:
> Hello @mstorsjo,
> 
> I was trying to build a C/C++ project a while ago (not sure whether it was ZLib or LLVM, but pretty sure it was a CMake-based project), and CMake was failing on `mt.exe`. However, if `mt.exe` ran again after failing the first time, CMake would pass.
> 
> I was about to make the following change to the script, but was hesitant whether or not this is a hacky fix:
> 
> ```
> diff --git a/wrappers/mt b/wrappers/mt
> index 30349ed..f6f359f 100755
> --- a/wrappers/mt
> +++ b/wrappers/mt
> @@ -1,3 +1,7 @@
>  #!/bin/bash
>  . $(dirname $0)/msvcenv.sh
>  $(dirname $0)/wine-msvc.sh $SDKBINDIR/mt.exe "$@"
> +
> +if [ $? -ne 0 ]; then
> +    $(dirname $0)/wine-msvc.sh $SDKBINDIR/mt.exe "$@"
> +fi
> ```

More context:

This was used with [cccl](https://github.com/swig/cccl), but I believe this issue will also happen if I use a CMake project in Linux that cross-compiles to MSVC Windows using this toolchain with `cccl`.

mstorsjo:
> > Hello @mstorsjo,
[...]

PR #63, opened 2023-04-14
Handle /notify_update in mt wrapper

huangqinjin:
CMake invokes mt.exe with the undocumented option, and mt.exe may exit with 0x41020001.

The exit code is truncated to 8 bits on POSIX hosts.

Translate exit code of mt.exe in bat file.

https://gitlab.kitware.com/cmake/cmake/-/blob/0991023c30ed5b83bcb1446b5bcc9c1eae028835/Source/cmcmd.cxx#L2533
https://gitlab.kitware.com/cmake/cmake/-/blob/0991023c30ed5b83bcb1446b5bcc9c1eae028835/Source/cmcmd.cxx#L2388

Fixes https://github.com/mstorsjo/msvc-wine/issues/45

### DIFF
--- a/wrappers/mt
+++ b/wrappers/mt
@@ -15,9 +15,4 @@
 # OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
 . $(dirname $0)/msvcenv.sh
-$(dirname $0)/wine-msvc.sh $SDKBINDIR/mt.exe "${@#/notify_update}"
-
-if [[ $? == 0 && "$@" =~ '/notify_update' ]]; then
-    # https://gitlab.kitware.com/cmake/cmake/-/blob/0991023c30ed5b83bcb1446b5bcc9c1eae028835/Source/cmcmd.cxx#L2388
-    exit 187
-fi
+$(dirname $0)/wine-msvc.sh $SDKBINDIR/mt.exe "$@"
--- a/wrappers/mt
+++ b/wrappers/mt
@@ -15,4 +15,9 @@
 # OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
 . $(dirname $0)/msvcenv.sh
-$(dirname $0)/wine-msvc.sh $SDKBINDIR/mt.exe "$@"
+$(dirname $0)/wine-msvc.sh $SDKBINDIR/mt.exe "${@#/notify_update}"
+
+if [[ $? == 0 && "$@" =~ '/notify_update' ]]; then
+    # https://gitlab.kitware.com/cmake/cmake/-/blob/0991023c30ed5b83bcb1446b5bcc9c1eae028835/Source/cmcmd.cxx#L2388
+    exit 187
+fi
--- a/wrappers/wine-msvc.bat
+++ b/wrappers/wine-msvc.bat
@@ -1,3 +1,8 @@
 @echo off
 
 %* %WINE_MSVC_ARGS% >%WINE_MSVC_STDOUT% 2>%WINE_MSVC_STDERR%
+
+REM https://gitlab.kitware.com/cmake/cmake/-/blob/0991023c30ed5b83bcb1446b5bcc9c1eae028835/Source/cmcmd.cxx#L2388
+if /I "%~n1"=="mt" (
+    if %errorlevel%==1090650113 (exit 187)
+)


mstorsjo:
If this patch is tested to work (I haven't tried the syntax `${var#pattern}` on something like `$@`), then this looks reasonable to me!

Do you happen to know in which cases cmake does try to execute `mt.exe`, i.e. if we'd add a minimal testcase that triggers both this issue and the one relating to `CMAKE_NINJA_CMCLDEPS_RC`, are we sure that we're triggering this? If we have cmake working without any custom modifications, we should add some CI coverage for it (see https://github.com/mstorsjo/msvc-wine/blob/master/.github/workflows/build.yml - currently I'm only compiling the bundled `hello.c` with `cl.exe` and `clang-cl`).

On the other hand, it looks like `mt.exe` is executed much less since https://gitlab.kitware.com/cmake/cmake/-/commit/0b552eb877b887638e8130bb6c982106a76827d8 anyway, so at some point we're probably losing the test coverage for it (unless we explicitly enable incremental linking, but that's probably not necessary).

huangqinjin:
If we silently ignore `/notify_update`, cmake does not execute `RC Pass 2:` and `FINAL LINK:`. So the final binary does not contain the manifest contents.

I reworked the PR to return `0xbb` in case `/notify_update` is present. Such that cmake can continue to execute `RC Pass 2:` and `FINAL LINK:`.

I think we can add a CI test for this case. Checking `requestedExecutionLevel` literal in the final binary is enough.

mstorsjo:
> If we silently ignore `/notify_update`, cmake does not execute `RC Pass 2:` and `FINAL LINK:`. So the final binary does not contain the manifest contents.

Oh, I see - so in builds with the patched cmake, it has actually never merged the manifest in the end?

> I reworked the PR to return `0xbb` in case `/notify_update` is present. Such that cmake can continue to execute `RC Pass 2:` and `FINAL LINK:`. Though sometimes these two steps are unneeded, but correctness first.

Interesting; good that this can be done with a return code within 0-255 and not requiring a full 32 bit windows style return code too.

> I think we can add a CI test for this case. Checking `requestedExecutionLevel` literal in the final binary is enough.

Ok, good. FWIW, for CI I don't think we strictly need to check that the manifest actually is included - I would settle for the level of smoke testing to see that building doesn't fail - we'd just need to build a test project that actually trigger including a manifest (do all cmake builds include that)?

huangqinjin:
> in builds with the patched cmake, it has actually never merged the manifest in the end?

No,  merging is always done, but the output  `embed.manifest` (may be updated!) is not embeded into binary this time linking. And when next time linking, `embed.manifest` is not deleted, and cmake directly embeds this file (already updated due to last `LINK Pass 1:` and `MT:`!). 
https://gitlab.kitware.com/cmake/cmake/-/blob/master/Source/cmcmd.cxx#L2441

That means if we don't handle `/notify_update`, cmake always embeds an old version manifest (empty for the first time linking). 

Worse, if there is a user provided manifest, and it is modified, cmake will re-link, but the changes is not included into final binary. (cmake supports manifests as source files, simply use `add_executable(exe main.cpp a.manifest)`).

> we'd just need to build a test project that actually trigger including a manifest (do all cmake builds include that)?

Yes, incremental link (default for debug build) always invokes `mt.exe`.

huangqinjin:
After further testing, `/notify_update` is supported but the exit code 0x41020001 is truncated to 1. So I added a batch file to translate 0x41020001 to 0xbb.

Testing:
```shell
> rm -f embed.manifest
> /opt/msvc/bin/x64/mt /nologo /manifest wmain.manifest /out:embed.manifest /notify_update
> echo $?
187
> /opt/msvc/bin/x64/mt /nologo /manifest wmain.manifest /out:embed.manifest /notify_update
> echo $?
0
```

 `wmain.manifest`.
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
  <application>
    <windowsSettings>
      <!-- https://docs.microsoft.com/en-us/windows/uwp/design/globalizing/use-utf8-code-page -->
      <activeCodePage xmlns="http://schemas.microsoft.com/SMI/2019/WindowsSettings">UTF-8</activeCodePage>
    </windowsSettings>
  </application>
</assembly>
```

huangqinjin:
`mt.bat` should be merged into `wine-msvc.bat` introduced in PR #65 .

mstorsjo:
LGTM, thanks!

PR #67, opened 2023-05-09
Fix bat file arguments handling

huangqinjin:
1. Escape all double-quotes.
2. EnableDelayedExpansion for other special characters: %^&|<>

Exclamation mark ! is not handled correctly by wine cmd currently.
https://bugs.winehq.org/show_bug.cgi?id=54928


mstorsjo:
Thanks!

PR #70, opened 2023-06-04
Unixify paths for more cases

### LINKED ISSUE #69: Work with ccache: Problem with MSVC /P (Preprocess to a File) option
### Background

`ccache` has [three working modes](https://ccache.dev/manual/4.8.1.html#_common_hashed_information):
- The preprocessor mode: Run preprocessor and parse the output file. It is slow.
- The direct mode: Fallback to preprocessor mode if it is not available. It is fast.
- The depend mode. Same as direct mode, but never fallback to preprocessor mode, and rely on outputs of `/showIncludes`.

### Problem
```shell
touch /tmp/header.h
echo '#include "header.h"' > /tmp/main.c
CCACHE_LOGFILE=ccache.log CCACHE_NODEPEND= CCACHE_DIRECT= ccache /opt/msvc/bin/x64/cl.exe -c /tmp/main.c
cat ccache.log
```

```
] Trying direct lookup
] Manifest key: 74199n1cl3ceh887hu1o509m3thr8mql2
] No 74199n1cl3ceh887hu1o509m3thr8mql2 in local storage
] Running preprocessor
] Executing /opt/msvc/bin/x64/cl.exe -P -Fi/run/user/1000/ccache-tmp/cpp_stdout.tmp.yFZLHy.i /tmp/main.c
] Failed to stat z:/tmp/main.c: No such file or directory
] Disabling direct mode
```

The direct mode is always disabled due to accessing `z:/tmp/main.c`.

### Cause
The output of `/opt/msvc/bin/x64/cl.exe -P /tmp/main.c` is the file `main.i` in CWD if option `/Fi` is not present.

```
#line 1 "z:/tmp/main.c"
#line 1 "z:\\tmp\\header.h"
#line 2 "z:/tmp/main.c"
```
The paths in the file are not converted to unix style. `ccache` will parse this file and hash the files listed, which is of cause failed, and then direct mode is disabled.

### Solution
- **Use ccache depend mode**. It doesn't run preprocessor but hash files listed in `/showIncludes`, which we already convert them to unix style paths. **OR**,
- Postprocess the preprocessor output file: convert to unix style paths. _This need implement in `wine-msvc.sh` wrapper_. @mstorsjo What's your opinion?

huangqinjin:
- Unixify paths for cl /showIncludes only. Otherwise outputs may be _partially_ converted, e.g.
  ```shell
  > echo '__FILE__;' > /tmp/test.c && /opt/msvc/bin/x64/cl /nologo /E /FC /tmp/test.c
  test.c
  #line 1 "\\tmp\\test.c"
  "\\tmp\\test.c";
  ```

- Unixify paths for dumpbin /PDBPATH also. Otherwise outputs were also _partially_ converted, e.g.
  ```
  > echo 'int main() {}' > /tmp/main.c && /opt/msvc/bin/x64/cl /nologo /Z7 /tmp/main.c /Fe/tmp/main.exe && /opt/msvc/bin/x64/dumpbin /nologo /PDBPATH /tmp/main.exe
  main.c

  Dump of file \tmp\main.exe

  File Type: EXECUTABLE IMAGE
    PDB file found at '\tmp\main.pdb'
  ```

  Vcpkg use it to implement the script `vcpkg_copy_pdbs`: https://github.com/microsoft/vcpkg/blob/662dbb50e63af15baa2909b7eac5b1b87e86a0aa/scripts/cmake/vcpkg_copy_pdbs.cmake#L24

- Unixify paths for cl /P /Fi. There are three cases of `cl /P`:
  i: A output file is specified by `/Fi`. This PR handles this case and resolves #69 .
  ii. & ii: One or more source files are given to `cl /P` and an optional output directory can be specified by `/Fi` (default to CWD). `cl` will generate `{output directory}/{basename of a source file without extension}.i` for each source file. This PR does **NOT** address these cases.

### DIFF
--- a/wrappers/cl
+++ b/wrappers/cl
@@ -18,8 +18,10 @@
 
 # /showIncludes
 unixify_path='/^Note: including file: /{ s/z:([\\/])/\1/i; s,\\,/,g; }'
+# /E
+unixify_line='/^[[:blank:]]*#[[:blank:]]*line[[:blank:]]/{ s/z:([\\/])/\1/i; s,\\\\,/,g; }'
 
-export WINE_MSVC_STDOUT_SED="$unixify_path"
+export WINE_MSVC_STDOUT_SED="$unixify_path;$unixify_line"
 export WINE_MSVC_STDERR_SED="$unixify_path"
 
 $(dirname $0)/wine-msvc.sh $BINDIR/cl.exe "$@"
--- a/wrappers/cl
+++ b/wrappers/cl
@@ -15,4 +15,11 @@
 # OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
 . $(dirname $0)/msvcenv.sh
+
+# /showIncludes
+unixify_path='/^Note: including file: /{ s/z:([\\/])/\1/i; s,\\,/,g; }'
+
+export WINE_MSVC_STDOUT_SED="$unixify_path"
+export WINE_MSVC_STDERR_SED="$unixify_path"
+
 $(dirname $0)/wine-msvc.sh $BINDIR/cl.exe "$@"
--- a/wrappers/cl
+++ b/wrappers/cl
@@ -25,3 +25,26 @@ export WINE_MSVC_STDOUT_SED="$unixify_path;$unixify_line"
 export WINE_MSVC_STDERR_SED="$unixify_path"
 
 $(dirname $0)/wine-msvc.sh $BINDIR/cl.exe "$@"
+
+ec=$?
+[ $ec -ne 0 ] && exit $ec
+
+# Postprocess
+for a in "$@"; do
+    case $a in
+        [-/]P) arg_P=$a ;;
+        [-/]Fi*) arg_Fi=${a:3} ;;
+    esac
+done
+
+# Unixify paths for /P
+if [ -n "$arg_P" ] && [ -f "$arg_Fi" ]; then
+    if sed --help 2>&1 | grep '\-i extension' >/dev/null; then
+        inplace=(-i '') # BSD sed
+    else
+        inplace=(-i)    # GNU sed
+    fi
+    sed "${inplace[@]}" -E 's/\r//;'"$unixify_line" "$arg_Fi"
+fi
+
+exit $ec
--- a/wrappers/wine-msvc.sh
+++ b/wrappers/wine-msvc.sh
@@ -46,10 +46,11 @@ while [ $# -gt 0 ]; do
 	shift
 done
 
-unixify_path='s/\r// ; s/z:\([\\/]\)/\1/i ; /^Note:/s,\\,/,g'
+WINE_MSVC_STDOUT_SED='s/\r//;'"$WINE_MSVC_STDOUT_SED"
+WINE_MSVC_STDERR_SED='s/\r//;'"$WINE_MSVC_STDERR_SED"
 
 if ! $HAS_MSVCTRICKS; then
-	WINEDEBUG=-all wine64 "$EXE" "${ARGS[@]}" 2> >(sed -e "$unixify_path" >&2) | sed -e "$unixify_path"
+	WINEDEBUG=-all wine64 "$EXE" "${ARGS[@]}" 2> >(sed -E "$WINE_MSVC_STDERR_SED" >&2) | sed -E "$WINE_MSVC_STDOUT_SED"
 	exit $PIPESTATUS
 else
 	export WINE_MSVC_STDOUT=${TMPDIR:-/tmp}/wine-msvc.stdout.$$
@@ -64,7 +65,7 @@ else
 
 	cleanup && mkfifo $WINE_MSVC_STDOUT $WINE_MSVC_STDERR || exit 1
 
-	sed -e "$unixify_path" <$WINE_MSVC_STDOUT &
-	sed -e "$unixify_path" <$WINE_MSVC_STDERR >&2 &
+	sed -E "$WINE_MSVC_STDOUT_SED" <$WINE_MSVC_STDOUT &
+	sed -E "$WINE_MSVC_STDERR_SED" <$WINE_MSVC_STDERR >&2 &
 	WINEDEBUG=-all wine64 "$EXE" "${ARGS[@]}" &>/dev/null
 fi
--- a/test/test-cl.sh
+++ b/test/test-cl.sh
@@ -51,4 +51,14 @@ Note: including file: ${CWD}test.h
 EOF
 
 
+EXEC cl-P-Fi ${BIN}cl /nologo /P /Ficl-P-Fi.i ./test.c
+DIFF cl-P-Fi.i - <<EOF
+#line 1 "./test.c"
+#line 1 "${CWD}test.h"
+#line 2 "./test.c"
+ 	 #line 5 "./test.c"
+const char* file = "./test.c";
+EOF
+
+
 EXIT
--- a/test/test-cl.sh
+++ b/test/test-cl.sh
@@ -22,6 +22,8 @@ EOF
 
 cat >test.c <<EOF
 #include "test.h"
+ 	 #line 5 __FILE__
+const char* file = __FILE__;
 EOF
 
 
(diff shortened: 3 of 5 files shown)


huangqinjin:
@mstorsjo It is worth unit testing these changes. Could you please add a basic test for `/showIncludes`? And then I will complete the tests.

mstorsjo:
(Sorry I don't have time to actually review this one right now, I'll have to get back to it later.)

> It is worth unit testing these changes. Could you please add a basic test for `/showIncludes`? And then I will complete the tests.

Hmm, what would be a good kind of test here? The practical tests would be of the form to just build something with e.g. ninja, and then rebuilding and making sure that it works (doesn't error out due to references to unknown paths), doesn't rebuild things needlessly, and actually does rebuild files as necessary if e.g. touching a header. But many of these aspects are hard and messy to verify in a test...

I guess what you have in mind is to manually run `cl /showIncludes` and inspect the output from it, and verify some properties of it? I'll see what I can do wrt that - I can hopefully give that a try within a few days.

mstorsjo:
> > manually run cl /showIncludes and inspect the output from it
> 
> Yes, this is what in my mind. For example, run
> 
> ```
> $ touch /tmp/header.h && echo '#include "header.h"' >/tmp/test.c && /opt/msvc/bin/x64/cl /nologo /showIncludes /c /tmp/test.c
> test.c
> Note: including file: /tmp/header.h
> ```
> 
> Save the output to a file as reference. The test just runs the same command, captures the outputs, compares to references byte-wise.

Ok, does db57d180ce54ce837470fa8dd6c133dfd574f6b6 seem reasonable?

wrappers/wine-msvc.sh
```diff
@@ -80,4 +81,28 @@ else
...
+	ec=$?
 fi
+
+[ $ec -ne 0 ] && exit $ec
+
+# Postprocess
+for a in "${ARGS[@]}"; do
+	case $a in
+		[-/]P) arg_P=$a ;;
+		[-/]Fi[zZ]:*) a=${a:5}; arg_Fi=${a//\\//} ;;
```
mstorsjo, line 93: This had me confused for a while, as I didn't expect callers to be calling us with arguments like `/Fiz:/home...`, but then I realize this runs on the translated arguments in `$ARGS`, so `z:/...` is indeed to be expected for absolute paths.

Would it make more sense to do this interpretation on the original unaltered arguments from before? I guess it doesn't matter much since it's easy to match these cases anyway.

Also, for the expansion `arg_Fi=${a//\\//}`, I don't quite see exactly which bash parameter substitution mechanism that uses - I presume it is related to translating backslashes to forward slashes?

Do we expect that to happen? I would expect callers to only call us with unix style paths here, and the translation above only injects `z:` but shouldn't convert to backslashes, right?

mstorsjo:
> Ok, does [db57d18](https://github.com/mstorsjo/msvc-wine/commit/db57d180ce54ce837470fa8dd6c133dfd574f6b6) seem reasonable?

I realized we definitely should run the same kind of test on macOS too. There are too many subtle differences in how `sed` behaves between GNU and (old) BSD tools...

wrappers/wine-msvc.sh
```diff
@@ -80,4 +81,28 @@ else
...
+		[-/]Fi*) a=${a:3}; arg_Fi=${a//\\//} ;;
+	esac
+done
+
+# Unixify paths for cl /P
+if [ -n "$arg_P" ] && [ -f "$arg_Fi" ]; then
+	if sed --help 2>&1 | grep '\-i extension' >/dev/null; then
```
mstorsjo, line 100: Oh, I hadn't expected that way of handling this. I guess this works, although it feels somewhat brittle.

Personally I just tend to avoid the `-i` form due to this ambiguity; I just pipe to a temp file and rename it afterwards. Not very elegant, but robust at least.

In any case, either form should be fine as long as we keep testing both platforms.

wrappers/wine-msvc.sh
huangqinjin: > I would expect callers to only call us with unix style paths here

Yes, this makes the expansion unnecessary.

Actually `/...` is also supported by wine/windows (relative to current drive). Did you encounter any issue without the translation to `z:/...`?

mstorsjo: I believe it was mostly needed to disambiguate paths from options, e.g. `/home/foo` can be interpreted as the option `/h` with argument `ome/foo`, while `z:/home/foo` makes it clear that it is a path, not an argument.

huangqinjin: OK, fair enough.

wrappers/wine-msvc.sh
huangqinjin: I like `sed -i` for the reason that it supports multiple files.

`/P` may outputs multiple `.i` files if multiple source files are provided. In this case we can simply replace `arg_Fi` with an array. Though this PR doesn't handle this case.

mstorsjo:
> > run the same kind of test on macOS too
> 
> Yes. I think we should write some source files, some reference files, and some scripts to test it.

Yes, that sounds reasonable - feel free to try to build something such!

huangqinjin:
I am thinking that, we can move the postprocesses to their corresponding wrappers. And export two environment variables `WINE_MSVC_STDOUT_SED` and `WINE_MSVC_STDERR_SED` from each wrapper and use them in the sed commands in `wine-msvc.sh`.

mstorsjo:
> I am thinking that, we can move the postprocesses to their corresponding wrappers. And export two environment variables `WINE_MSVC_STDOUT_SED` and `WINE_MSVC_STDERR_SED` from each wrapper and use them in the sed commands in `wine-msvc.sh`.

I guess that sounds reasonable.

wrappers/wine-msvc.sh
```diff
@@ -46,10 +46,11 @@ while [ $# -gt 0 ]; do
 	shift
 done
 
-unixify_path='s/\r// ; s/z:\([\\/]\)/\1/i ; /^Note:/s,\\,/,g'
+WINE_MSVC_STDOUT_SED='s/\r//;'"$WINE_MSVC_STDOUT_SED"
```
mstorsjo, line 49: For the frontend scripts that don't set `WINE_MSVC_STDOUT_SED`, we end up with a trailing `;` here without any extra regex following it; is that ok, or should we only add the `;` if `WINE_MSVC_STDOUT_SED` is nonempty here?

We should probably add similar tests for e.g. ml64 and armasm too. I can look into tests for the arm tools at least.

wrappers/cl
```diff
@@ -18,8 +18,10 @@
 
 # /showIncludes
 unixify_path='/^Note: including file: /{ s/z:([\\/])/\1/i; s,\\,/,g; }'
+# /E
+unixify_line='/^[[:blank:]]*#line[[:blank:]]/{ s/z:([\\/])/\1/i; s,\\\\,/,g; }'
```
mstorsjo, line 22: Should we tolerate blanks between `#` and `line` too? I guess MSVC doesn't produce that, but just as `#   define FOO` is somewhat common as indentation, I guess `#  line` also could be considered accepted? (Likewise, I guess we could accept `[[:blank:]]+` after `line`?)

wrappers/cl
```diff
@@ -25,3 +25,26 @@ export WINE_MSVC_STDOUT_SED="$unixify_path;$unixify_line"
...
+if [ -n "$arg_P" ] && [ -f "$arg_Fi" ]; then
+    if sed --help 2>&1 | grep '\-i extension' >/dev/null; then
+        inplace=(-i '') # BSD sed
+    else
+        inplace=(-i)    # GNU sed
+    fi
+    sed "${inplace[@]}" -E 's/\r//;'"$unixify_line" "$arg_Fi"
+fi
+
+exit $ec
```
mstorsjo, line 50: I guess at this point, `$ec` can only be zero?

mstorsjo (commented):
This looks really good to me now, thanks! Just a couple details for discussion.

wrappers/cl
huangqinjin: I actually tested this. The preprocessor compresses all blank characters in `#    line    5    "a.c"`, and becomes `#line 5 "a.c"`. But blank characters before `#` are preserved.

mstorsjo: Ok, good to know. I meant more from the point of view of being flexible in accepting what MSVC might output, i.e. it could change to not normalize those lines in that way, in a future update. I.e., not only matching exactly what we get today, but matching anything that semantically means the same thing.

huangqinjin: ok, fair enough.

wrappers/cl
mstorsjo: Right, that's a good question - I'm not entirely decided which way is better, but I don't think it matters very much either. If the sed commands are broken, we need to know about it though, and with the tests we should be able to catch it there at least.

wrappers/cl
mstorsjo: Oh, right, that's true, as we're not covering the file name itself with the RE, only selecting which lines to apply it on.

wrappers/cl
huangqinjin: > I don't think it matters very much either

So do I.

wrappers/wine-msvc.sh
huangqinjin: I think trailing `;` is ok, otherwise we had already observed some errors from `rc`.

huangqinjin: I don't have usecases for other tools, no idea what kind of tests should be added now.

mstorsjo:
LGTM, thanks!

PR #74, opened 2023-06-10
Refactor tests

huangqinjin:
@mstorsjo Please feel free to push more commits on this PR.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.13.4)
+cmake_minimum_required(VERSION 3.25)
 project(hello)
 
 set(hello_SOURCES hello.c)
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -6,3 +6,9 @@ if (WIN32)
     list(APPEND hello_SOURCES hello.rc)
 endif()
 add_executable(hello ${hello_SOURCES})
+
+
+if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" AND DEFINED MSVCDIR)
+    add_library(cxxmodules STATIC "${MSVCDIR}/modules/std.ixx")
+    target_compile_options(cxxmodules PUBLIC /std:c++latest)
+endif()
--- a/test/test-cmake.sh
+++ b/test/test-cmake.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+#
+# Copyright (c) 2023 Huang Qinjin
+#
+# Permission to use, copy, modify, and/or distribute this software for any
+# purpose with or without fee is hereby granted, provided that the above
+# copyright notice and this permission notice appear in all copies.
+#
+# THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+# WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+# ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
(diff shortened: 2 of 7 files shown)


test/test-cmake.sh
```diff
@@ -0,0 +1,45 @@
...
+    -DCMAKE_BUILD_TYPE=RelWithDebInfo
+    -DCMAKE_SYSTEM_NAME=Windows
+)
+
+case $OSTYPE in
+    darwin*)
+        CMAKE_ARGS+=(
+            -DCMAKE_EXE_LINKER_FLAGS=/MANIFEST:NO
+            -DCMAKE_SHARED_LINKER_FLAGS=/MANIFEST:NO
+            -DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded
```
mstorsjo, line 30: Why are these arguments needed on Darwin? I presume that means that msvctricks.exe doesn't work there (breaking the mt.exe return code rewrite, and leaving cmake/ninja waiting on the mspdbsrv background process)? A comment would probably be good.

mstorsjo:
I think this looks mostly reasonable. The recursive use of `test.sh` is a bit mind-twisting at first, but I think I see the reasons for it.

test/test-cmake.sh
huangqinjin: https://github.com/mstorsjo/msvc-wine/pull/65#issuecomment-1531747302

mstorsjo: Oh, ok, thanks for reminding me! A comment mentioning these reasons would be good.

huangqinjin: Done.

test/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.13.4)
+cmake_minimum_required(VERSION 3.25)
 project(hello)
 
 set(hello_SOURCES hello.c)
```
huangqinjin, line 4: @mstorsjo Is the `WIN32` condition necessary? Is this cmakelist supposed to target platforms other than Windows?

mstorsjo: I'd like the test to remain compileable for any platform, yeah. It's not something that we'd do much, admittedly, but I'd like the cmakelist to remain like a real one would be for a proper cross platform project.

huangqinjin:
It seems there is a problem with Cmake when targeting ARM, though it doesn't fail.

https://github.com/mstorsjo/msvc-wine/actions/runs/5269443375/jobs/9527585853#step:5:129

mstorsjo:
> It seems there is a problem with Cmake when targeting ARM, though it doesn't fail.
> 
> https://github.com/mstorsjo/msvc-wine/actions/runs/5269443375/jobs/9527585853#step:5:129

It seems like CMake doesn't pass `/EHsc` when compiling C++ code targeting ARM, contrary to other architectures.

mstorsjo:
> I think this PR is ready and if it looks good to you, please **squash and merge**.

It looks reasonable to me, thanks! I pushed a minor cosmetic update to the branch. Although to me, the commits look neatly separated, so to me I'd actually prefer to just do a rebase+merge here, if you don't mind.

huangqinjin:
> prefer to just do a rebase+merge here

I'm OK with that.

test/test-cmake.sh
```diff
@@ -0,0 +1,47 @@
...
+        ) ;;
+esac
+
+EXEC "" CC=${BIN}cl CXX=${BIN}cl RC=${BIN}rc cmake -S"$TESTS" -GNinja "${CMAKE_ARGS[@]}"
+EXEC "" ninja -v
+
+# Rerun ninja to make sure that dependencies aren't broken.
+EXEC ninja-rerun ninja -d explain -v
+DIFF ninja-rerun.out - <<EOF
+ninja: no work to do.
```
mstorsjo, line 48: If this testcase fails, the explanation is stored in `ninja-rerun.err`, but that's not printed anywhere. It'd be nice if that was printed, especially if the diff test failed.

A second note; the diff is a bit hard to read when it's reversed (the actual output is the original state and the expected output is the new state, in the diff); it's easy to change by swapping the sides in the diff into `DIFF - ninja-rerun.out`, but that makes it slightly harder to read when the output name doesn't line up with the `EXEC` line above.

PR #75, opened 2023-06-15
Convert clang-cl tests to the test scripts, add tests for tricky header casing

mstorsjo:
CC @huangqinjin

### DIFF
--- a/test/headers.cpp
+++ b/test/headers.cpp
@@ -1,0 +1,23 @@
+// These attributes pick a different implementation within sal.h,
+// that include CodeAnalysis/sourceannotations.h in sal.h, checking that
+// such paths with originally mixed casing works.
+// The attributes based implementation in sal.h doesn't work with clang-cl,
+// only with MSVC.
+//
+// Enabling these codepaths makes Wine print the following messages while
+// executing cl.exe:
+// 070c:err:winediag:nodrv_CreateWindow Application tried to create a window, but no driver could be loaded.
+// 070c:err:winediag:nodrv_CreateWindow L"The explorer process failed to start."
+// 070c:err:systray:initialize_systray Could not create tray window
+#define _USE_DECLSPECS_FOR_SAL 0
+#define _USE_ATTRIBUTES_FOR_SAL 1
+#include <windows.h>
+#include <sal.h>
+
+#include <GL/gl.h>
+
+// comdef.h can only be included in C++ mode. It includes Ole2.h and OleCtl.h
+// which exist with a different casing in the WinSDK.
+// This includes MSVC STL headers, which strictly require a very recent
+// version of Clang.
+#include <comdef.h>
--- a/test/test-clang-cl-cmds.sh
+++ b/test/test-clang-cl-cmds.sh
@@ -37,4 +37,10 @@ EXEC "" clang-cl --target=$TARGET_ARCH-windows-msvc "${TESTS}hello.c" -Fehello.e
 EXEC "" clang-cl --target=$TARGET_TRIPLE "${TESTS}hello.c" -c -Fohello.obj
 EXEC "" lld-link hello.obj -out:hello.exe
 
+# Test if paths with potentially problematic casing can be included. As this
+# is tricky to compile (the old sal.h codepaths that are enabled aren't
+# supported by Clang, and MSVC STL headers require a very recent Clang), just
+# test preprocessing it to make sure that all paths can be found.
+EXEC "" clang-cl --target=$TARGET_TRIPLE "${TESTS}headers.cpp" -P -Fiheaders-preproc.cpp
+
 EXIT
--- a/test/test-clang-cl-cmds.sh
+++ b/test/test-clang-cl-cmds.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+#
+# Copyright (c) 2023 Martin Storsjo
+#
+# Permission to use, copy, modify, and/or distribute this software for any
+# purpose with or without fee is hereby granted, provided that the above
+# copyright notice and this permission notice appear in all copies.
+#
+# THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+# WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+# ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+# WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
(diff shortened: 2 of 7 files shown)


test/test-cmake-clang-cl.sh
```diff
@@ -0,0 +1,50 @@
...
+fi
+
+CMAKE_ARGS=(
+    -DCMAKE_BUILD_TYPE=RelWithDebInfo
+    -DCMAKE_SYSTEM_NAME=Windows
+    -DCMAKE_MT=$(which llvm-mt)
+)
+
+# Using EXEC here breaks the quoted variables such as CC="..."
+CC="clang-cl --target=$ARCH-windows-msvc" CXX="clang-cl --target=$ARCH-windows-msvc" RC="llvm-rc" cmake -S"$TESTS" -GNinja "${CMAKE_ARGS[@]}"
```
mstorsjo, line 40: Note, here I had to skip using `EXEC` as the `eval` in it breaks passing the quoted `CC="..."` variables.

test/test-clang-cl.sh
```diff
@@ -0,0 +1,40 @@
...
+# WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+# ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+# WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+# ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+# OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+. "${0%/*}/test.sh"
+
+cd "$TESTS"
```
mstorsjo, line 19: Since this isn't `test.sh` itself but sources it, it creates an unnecessary temp directory and we manually have to `cd` into `$TESTS`, but that's probably benign and not worth making too much of a mess out of.

test/CMakeLists.txt
```diff
@@ -8,6 +8,23 @@ endif()
...
+    # once will succeed, as MSVC in Wine treats paths case insensitively.
+    # But a case sensitive dependency tracker in Ninja will force unnecessary
+    # recompilation, if the files don't exist with the right casing.
+    #
+    # This is tested only with MSVC, not with clang-cl, as MSVC STL is
+    # strict with what version of Clang it supports (currently requiring
+    # Clang 15, while common repositories only provide Clang 14). The
+    # header also triggers using a form of attributes for SAL that Clang
+    # doesn't support.
+    add_library(test_headers STATIC headers.cpp)
```
mstorsjo, line 23: I'm not sure if it's worth adding this testcase to the cmake file; the case sensitive clang-cl test in the previous commit should give quite enough coverage here. This makes the cmake test quite a bit slower, as cl.exe seems to be doing something weird when we compile this file (due to enabling the old style attributes for sal.h).

test/test-clang-cl.sh
```diff
@@ -0,0 +1,40 @@
...
+
+MSVC="${1:-/opt/msvc}"
+
+for arch in i686 x86_64 armv7 aarch64; do
+    case $arch in
+    i*86) msvcarch=x86 ;;
+    x86_64) msvcarch=x64 ;;
+    arm*) msvcarch=arm ;;
+    aarch64) msvcarch=arm64 ;;
+    esac
```
huangqinjin, line 29: I thinks this block can be moved to `msvcenv-native.sh` as it is also needed for users. And the logic should be reversed: construct clang triple from `ARCH` (sourced from `msvcenv.sh`).

test/test-clang-cl.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+#
+# Copyright (c) 2023 Huang Qinjin
```
huangqinjin, line 3: Your name.

test/test-clang-cl.sh
mstorsjo: Oh, that's indeed a good idea. We could probably even use `$ARCH` right away as such from there, but could also make `msvcenv-native.sh` produce a full triple.

Anyway, that doesn't help for the cases when using `/winsysroot`, before sourcing `msvcenv-native.sh` though. But I guess we can reverse this and primarily use the MSVC arch names here, and then map this as necessary within `test-clang-cl-cmds.sh`.

test/test-clang-cl.sh
mstorsjo: Ok, fair enough. These files were mostly boilerplate from your creations anyway, but I can change it in the newly created files.

mstorsjo:
I pushed updates now, taking advantage of the improvements to the scripts as you made, and applying your feedback. Thanks!

test/test-clang-cl-cmds.sh
```diff
@@ -0,0 +1,46 @@
...
+# pointing at the installation.
+. ${TESTS}../msvcenv-native.sh
+EXEC "" clang-cl --target=$TARGET_TRIPLE "${TESTS}hello.c" -c -Fohello.obj
+EXEC "" lld-link hello.obj -out:hello.exe
+
+# Test if paths with potentially problematic casing can be included. As this
+# is tricky to compile (the old sal.h codepaths that are enabled aren't
+# supported by Clang, and MSVC STL headers require a very recent Clang), just
+# test preprocessing it to make sure that all paths can be found.
+EXEC "" clang-cl --target=$TARGET_TRIPLE "${TESTS}headers.cpp" -E -P -Fiheaders-preproc.cpp
```
huangqinjin, line 44: `-E` is redundant.

test/headers.cpp
```diff
@@ -0,0 +1,21 @@
...
+// only with MSVC.
+//
+// Enabling these codepaths makes Wine print the following messages while
+// executing cl.exe:
+// 070c:err:winediag:nodrv_CreateWindow Application tried to create a window, but no driver could be loaded.
+// 070c:err:winediag:nodrv_CreateWindow L"The explorer process failed to start."
+// 070c:err:systray:initialize_systray Could not create tray window
+#define _USE_DECLSPECS_FOR_SAL 0
+#define _USE_ATTRIBUTES_FOR_SAL 1
+#include <windows.h>
```
huangqinjin, line 14: `<GL/gl.h>` is also a good test.

test/test-clang-cl-cmds.sh
mstorsjo: Ok, removed the `-E`.

test/headers.cpp
mstorsjo: Thanks, added `<GL/gl.h>` into the test.

mstorsjo:
FWIW I'll probably leave out the `headers.cpp` test from the cmake file (leaving out the last commit before merging, if the rest is ok with you) as it is a bit messy and doesn't add much extra test coverage.

huangqinjin:
> leave out the headers.cpp test from the cmake file

I second this. And LGTM!

PR #76, opened 2023-06-15
Improve test scripts

huangqinjin:


huangqinjin:
`env` supports zero `k=v` pair. The problem is it must execute a true file existing on file system. It's reasonable as it is not a shell builtin. `test-mt` executes `mtRetIsUpdate` which is a bash function. So `env` cannot be used here.

Actually escaping is better since we echo the escaped command, now we can directly copy and paste into terminal in case we want to run it manually.

PR #78, opened 2023-07-05
wrappers: Document adding z: prefixes on absolute paths, add z: prefixes in one more case

mstorsjo:
CC @huangqinjin 

This helps fixing compiling LLVM with msvc-wine with recent Wine versions (Wine 8.1 or newer affected) - see https://bugs.winehq.org/show_bug.cgi?id=55200 for details.

### DIFF
--- a/wrappers/wine-msvc.sh
+++ b/wrappers/wine-msvc.sh
@@ -27,6 +27,19 @@ ARGS=()
 while [ $# -gt 0 ]; do
 	a=$1
 	case $a in
+	[-/][A-Za-z]/*)
+		opt=${a:0:2}
+		path=${a:2}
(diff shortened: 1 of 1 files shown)


huangqinjin:
Passing relative paths to `-I` has same issue. `ccache` converts absolute paths to relative paths if `CCACHE_BASEDIR` is set. But I haven't encounter the issue yet. And `-external:I` option also has same issue.

I would like any tricks if it can solve real project issues, even though it is not perfect or complete. So LGTM.

mstorsjo:
> Passing relative paths to `-I` has same issue. `ccache` converts absolute paths to relative paths if `CCACHE_BASEDIR` is set. But I haven't encounter the issue yet. And `-external:I` option also has same issue.

Oh, I hadn't seen that option used. That option is a bit more of a mess to match...

PR #79, opened 2023-07-05
test: Add testing of the arch specific assembler tools

mstorsjo:
Just testing that running the assembler exits successfully.

Adding the testing in test-cl.sh, instead of a separate test script, although I'm open to creating a separate test script if that's preferred.

CC @huangqinjin

### DIFF
--- a/test/test-asm.sh
+++ b/test/test-asm.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+#
+# Copyright (c) 2023 Martin Storsjo
+#
+# Permission to use, copy, modify, and/or distribute this software for any
+# purpose with or without fee is hereby granted, provided that the above
+# copyright notice and this permission notice appear in all copies.
+#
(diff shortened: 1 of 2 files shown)


huangqinjin:
LGTM though I would prefer a separate script for `masm`.

mstorsjo:
> LGTM though I would prefer a separate script for `masm`.

Do you mean specifically the x86 `masm` tests, or splitting all of the newly added ones here into a separate `test-asm.sh`?

huangqinjin:
> splitting all of the newly added ones here into a separate test-asm.sh

That is what I mean. Sorry for confusion.

mstorsjo:
Updated the branch to move the tests to a separate file. I also reordered the tests in `test.sh` to test the individual tools first before testing cmake/meson.

huangqinjin:
LGTM!

PR #80, opened 2023-07-05
test: Test building Meson based projects with msvc-wine

mstorsjo:
Only testing building with MSVC - not testing building with clang-cl with Meson. Clang-cl does work with Meson, but for the resource compiler, Meson currently fails to detect llvm-rc.

Therefore, currently if building with Meson, one has to use the Wine wrapped version of rc.exe, which requires Wine to be available, contrary to the rest of the clang-cl/lld based setup. (The CI environment for the clang/lld tests intentionally does not have Wine available.)

### DIFF
--- a/test/test-meson.sh
+++ b/test/test-meson.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+#
+# Copyright (c) 2023 Martin Storsjo
+#
+# Permission to use, copy, modify, and/or distribute this software for any
+# purpose with or without fee is hereby granted, provided that the above
+# copyright notice and this permission notice appear in all copies.
+#
+# THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+# WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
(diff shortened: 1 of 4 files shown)


mstorsjo:
CC @huangqinjin

huangqinjin:
LGTM!

test/test-meson.sh
```diff
@@ -0,0 +1,88 @@
...
+MESON_ARGS=(
+    --cross-file cross.txt
+)
+
+case $OSTYPE in
+    darwin*)
+        MESON_ARGS+=(
+            # No winbind package available on macOS.
+            # https://github.com/mstorsjo/msvc-wine/issues/6
+            --buildtype release
```
huangqinjin, line 68: It seems that Meson write PDB separately for each source file, as opposed to CMake, which writes PDB to same `vc140.pdb` per target. So `/FS` option and `mspdbsrv.exe` is not used in Meson, means that it should support debug build.

Meson: https://github.com/mstorsjo/msvc-wine/actions/runs/5468446610/jobs/9956102203#step:5:79
CMake: https://github.com/mstorsjo/msvc-wine/actions/runs/5468446610/jobs/9956102203#step:5:41

mstorsjo: Thanks for catching it, this should be simplified by #82.

PR #81, opened 2023-07-05
RFC: test: Add github actions tests for building real world projects with msvc-wine

mstorsjo:
CC @huangqinjin 

This integrates testing of three real world projects that can be compiled with msvc-wine; dav1d (using Meson), which builds in 4 minutes in total (including installing msvc-wine) on Github Actions, ffmpeg (using a custom make based build system), which builds in 30 minutes, and llvm (which uses cmake), which builds in 3h 30min.

Including the full llvm testcase is most probably overkill, but on the other hand, it serves as a great reference testcase. (It also manages to trigger certain rare bugs, see #78.) Ffmpeg with 30 minutes is also significantly slower than the rest of the test cases we have. Dav1d with 4 minutes is quite tolerable on the other hand, but I'm not sure how much extra test coverage it adds on top of #79 and #80.

Not sure what the best way forward would be here. Maybe including them in the github actions file, but disabled by default, so one can manually enable them to test running them in a test branch?

### DIFF
--- a/.github/workflows/projects.yml
+++ b/.github/workflows/projects.yml
@@ -1,0 +1,168 @@
+name: Test building external projects
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - '.github/workflows/projects.yml'
+      - 'install.sh'
+      - 'wrappers/*'
+      - 'fixinclude'
+      - 'lowercase'
+
+jobs:
+  test-build-dav1d:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install prerequisites
+        run: |
+          sudo apt-get update && sudo apt-get install wine64-development python3 msitools python3-simplejson python3-six ca-certificates ninja-build winbind meson nasm
+          wine64 wineboot
+      - uses: actions/checkout@v3
+      - name: Download MSVC
+        run: |
+          ./vsdownload.py --accept-license --dest $(pwd)/msvc
+          ./install.sh $(pwd)/msvc
+      - name: Checkout gas-preprocessor
+        uses: actions/checkout@v3
+        with:
+          repository: ffmpeg/gas-preprocessor
+          ref: 9309c67acb535ca6248f092e96131d8eb07eefc1
+          path: gas-preprocessor
+      - name: Add gas-preprocessor to path
+        run: |
+          echo $(pwd)/gas-preprocessor >> $GITHUB_PATH
+      - name: Checkout dav1d
+        run: |
+          git clone https://code.videolan.org/videolan/dav1d.git
+          cd dav1d
+          git checkout 1.2.1
+      - name: Build dav1d
+        run: |
+          cd dav1d
+          mkdir build
+          cd build
+          export PATH=$(pwd)/../../msvc/bin/arm64:$PATH
+          cat >cross.txt <<EOF
+          [binaries]
+          c = 'cl'
+          cpp = 'cl'
+          ar = 'lib'
+          windres = 'rc'
+          ;exe_wrapper = ['wine']
+
+          [properties]
+          needs_exe_wrapper = true
+
+          [host_machine]
+          system = 'windows'
+          cpu_family = 'aarch64'
+          cpu = 'aarch64'
+          endian = 'little'
+          EOF
+
+          meson setup --cross-file cross.txt --buildtype release --prefix $(pwd)/../../install ..
+          ninja -v
+          ninja install
+          cp ../../msvc/vc/Redist/MSVC/1*/arm64/*.CRT/vcruntime*.dll ../../install/bin
+      # The build output from MSVC actually is redistributable, even if the
+      # toolchain itself isn't.
+      - uses: actions/upload-artifact@v3
+        with:
+          name: dav1d
+          path: |
+            install/*
+          retention-days: 7
+
+  test-build-ffmpeg:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install prerequisites
+        run: |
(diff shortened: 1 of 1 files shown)


huangqinjin:
My thought is that, these real projects develop more rapidly than this project, so rather than pinning to some commits of these projects here, we can add `msvc-wine` branch to them (forked under your account) respectively. Once upstreams have formal releases, we merge them into `msvc-wine` and run CI. If there are any issues, we add minimal repro back to this repo.

mstorsjo:
> My thought is that, these real projects develop more rapidly than this project, so rather than pinning to some commits of these projects here, we can add `msvc-wine` branch to them (forked under your account) respectively. Once upstreams have formal releases, we merge them into `msvc-wine` and run CI.

Hmm - I don't see that much of a difference between having a pinned known-working version in the tests here (which can be updated once a year, if relevant), compared to maintaining a separate branch in a fork?

> If there are any issues, we add minimal repro back to this repo.

FWIW I do build dav1d and ffmpeg with msvc-wine (but with an old install of an older MSVC version) nightly and weekly on a local machine already. LLVM I only test build with msvc-wine sporadically though.

> Or is it possible to trigger CI of these projects using GitHub APIs once we update this repo?

My intent isn't so much to track whether changes further upstream break msvc-wine or find a new corner case that needs to be fixed (that happens very seldom, from experience with running such test builds for many years already).

My intent is more of a full litmus test of msvc-wine actually working correctly. Yes, ideally the set of minimal repros we have in the tests here should be enough to cover everything. But in practice, we know it isn't.

So from my point of view, by running these build tests, e.g. after tweaking parameter mangling in `wine-msvc.sh` like the many changes you did back and forth earlier (where the ffmpeg build showed issues we didn't catch in small tests), we could know with some amount of confidence that the msvc-wine wrappers still work for these large real world projects. Or for testing that e.g. the `z:` prefixing for options like `-Fo/absolute/path` currently seems to not be needed by any of these projects.

I.e. this could be an opt-in extra test we could run when we feel that it'd be useful; to be run while code is still on a branch, under review - not after merging things to master and only then realize we broke something.

I'll see if it'd work to have a manually triggered job with `workflow_dispatch`, that we could run when we feel we want a full test before merging something.

huangqinjin:
OK. Then I think we can enable these three projects unconditionally as we rarely update this repo. Maybe can move them into another workflow, so that we can apply more precise conditions (e.g. only some paths changed or exclude some paths) to trigger it.

mstorsjo:
Updated this to be a separate workflow in github actions, which can be triggered manually from the github actions UI (once merged to the master branch) - where one can specify the branch for which it should be run. (It's also easy to add `push:` to the .yml file in a testing branch to run these tests automatically.)

huangqinjin:
> where one can specify the branch for which it should be run

Can it run for PR?

mstorsjo:
> > where one can specify the branch for which it should be run
> 
> Can it run for PR?

With the setup currently in this branch, you'd manually trigger the action from github.com/<username>/<repo>/actions (but looking at your fork, it seems you've already succeeded to do that), whenever that feels relevant.

Alternatively, I guess we could add it to always run on pull requests, whenever they touch relevant files. E.g. something like this?
```
on:
  pull_request:
    paths:
      - 'install.sh'
      - 'wrappers/*'
```

huangqinjin:
> we could add it to always run on pull requests, whenever they touch relevant files

I second this. Several hours CI running is acceptable from my point of view.

mstorsjo:
> > we could add it to always run on pull requests, whenever they touch relevant files
> 
> I second this. Several hours CI running is acceptable from my point of view.

Ok, updated the PR with that.

huangqinjin:
Shall we also add `fixinclude` and `lowercase`, which are used when installing. 

It seems current PR doesn't trigger CI.

mstorsjo:
> Shall we also add `fixinclude` and `lowercase`, which are used when installing.

Sure, I can do that.

> It seems current PR doesn't trigger CI.

Indeed, as the PR doesn't touch any of those files. I can do a push of the PR with an extra temporary commit that triggers running it unconditionally, to make sure the final form of the job is fine (although I should have tested it already a couple times).

huangqinjin:
Can add `projects.yml` itself to the list? (shouldn't this be the default?)

mstorsjo:
> Can add `projects.yml` itself to the list? (shouldn't this be the default?)

Yeah, I came to think of that later too, as that'll make maintainance of the workflow itself easier. Updated to include that now.

.github/workflows/projects.yml
```diff
@@ -0,0 +1,168 @@
...
+          ref: n6.0
+          path: ffmpeg
+      - name: Build ffmpeg
+        run: |
+          mkdir ffmpeg-build
+          cd ffmpeg-build
+          export PATH=$(pwd)/../msvc/bin/arm64:$PATH
+          ../ffmpeg/configure --arch=arm64 --target-os=win32 --toolchain=msvc --enable-cross-compile --prefix=$(pwd)/../install
+          make -j$(nproc)
+          make install
```
huangqinjin, line 110: Unlike other two projects, you don't copy vc runtime dll here, is it because ffmpeg links to static runtime?

mstorsjo: Yes, that's correct - by default, ffmpeg doesn't add any `-MD` options or similar, so it defaults to what `cl.exe` defaults to, which is static linking of the runtime.

huangqinjin: ok.

mstorsjo:
@huangqinjin Any further thoughts about this, or should I just go ahead and merge it?

huangqinjin:
It looks good enough to me now!

PR #82, opened 2023-07-08
test: Remove a special case for Meson testing on macOS

mstorsjo:
Meson doesn't use the /FS option when generating PDB debug info, which means that it doesn't use mspdbsrv.exe. Thus, the default debug mode should work fine on macOS too.

### DIFF
--- a/test/test-meson.sh
+++ b/test/test-meson.sh
@@ -60,15 +60,6 @@ MESON_ARGS=(
     --cross-file cross.txt
 )
 
-case $OSTYPE in
-    darwin*)
-        MESON_ARGS+=(
-            # No winbind package available on macOS.
-            # https://github.com/mstorsjo/msvc-wine/issues/6
-            --buildtype release
-        ) ;;
-esac
-
 export PATH="$BIN:$PATH"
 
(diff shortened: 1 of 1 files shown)


huangqinjin:
LGTM!

huangqinjin:
Now it links to debug runtime libraries (`/MDd`), which may be trouble as [README said](https://github.com/mstorsjo/msvc-wine#can-it-build-debug-versions). So Adding `--buildtype debugoptimized` is better, which matches what `test-cmake` `RelWithDebInfo` does. Generating PDB in release mode is also the default behavior of Visual Studio.

mstorsjo:
> Now it links to debug runtime libraries (`/MDd`), which may be trouble as [README said](https://github.com/mstorsjo/msvc-wine#can-it-build-debug-versions). So Adding `--buildtype debugoptimized` is better, which matches what `test-cmake` `RelWithDebInfo` does. Generating PDB in release mode is also the default behavior of Visual Studio.

Ok, I added another commit on this PR, adding `--buildtype debugoptimized` to `test-meson.sh` when building on any OS.

huangqinjin:
LGTM now.

PR #89, opened 2023-08-24
Convert install scripts to python

AgentK9:
Python is more universal than perl and allows for better extensibility and portability, not to mention readability.


mstorsjo:
I'm somewhat ambivalent about this PR. (I haven't studied your suggested code in detail yet.)

Yes, python is probably to prefer over perl for the `fixinclude` and `lowercase` scripts. But for the `install.sh` script, I'm not really sure about whether this is a net win or loss - what do we gain from converting that to python?

(The actions run that I just enabled will probably fail; this is due to a tool regression that I've worked around on master; if you rebase your branch on latest master, it should run successfully again.)

AgentK9:
Hello,

Thanks so much for your time in reviewing this. I have merged main into my branch. If you'd like, I can also squash all of these commits.

Here are a few reasons why the `install.sh` script should be changed along with the others: 

1. Unified Codebase: Aligns with the existing Python components, streamlining maintenance.
2. Inclusivity: Facilitates contributions from Windows developers unfamiliar with GNU shell/sh.
3. Versatility: Enhances portability and allows for future feature integrations with ease.

I'd really appreciate any feedback you have.

SaifRushdHadad:
> I'm somewhat ambivalent about this PR. (I haven't studied your suggested code in detail yet.)

@mstorsjo, I was wondering if you had reviewed the suggested changes? I know it has been some time since the PR was submitted but rebasing and updating should be doable.

PR #101, opened 2024-01-14
Fortify WinSDK version detection in install.sh

ravenexp:
install.sh detects the WinSDK version by scanning the common ./kits/10/include/ directory, but other kits such as WDK add non-versioned directories there, for example ./kits/10/include/wdf/.

Use a more narrow glob expression to ignore these extra directories.

Remove a redundant SDK version variable assignment and print the detected SDK and MSCV versions for diagnostics.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -105,7 +105,9 @@ fi
 ln_s Lib lib
 ln_s Include include
 cd ../..
-SDKVER=$(basename $(echo kits/10/include/* | awk '{print $NF}'))
+
+SDKVER=$(basename $(echo kits/10/include/10.* | awk '{print $NF}'))
+echo Using SDK version $SDKVER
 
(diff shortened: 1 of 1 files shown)


mstorsjo:
Looks good to me overall, thanks! There's a typo in the commit message, `MSCV` that should be `MSVC` though, other than that this looks nice.

ravenexp:
> There's a typo in the commit message, `MSCV` that should be `MSVC` though, other than that this looks nice.

Fixed the commit message. Thanks for catching this.

mstorsjo:
> > There's a typo in the commit message, `MSCV` that should be `MSVC` though, other than that this looks nice.
> 
> Fixed the commit message. Thanks for catching this.

Thanks, I'll merge this sometime after the CI jobs complete.

PR #103, opened 2024-01-16
Install MSBuild.exe

### LINKED ISSUE #85: MSBuild
I'd like to figure out how to install MSBuild - even when manually specifying it, it seems to be skipped.

And thanks for this project!

ravenexp:
1. Install unpacked MSBuild files.
2. Add `msbuild.exe` wrapper script.
3. Add MSBuild test script and a test VS2022 project for it.

Resolves #85


wrappers/msbuild
```diff
@@ -0,0 +1,43 @@
...
+# WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+# ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+# WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+# ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+# OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+. $(dirname $0)/msvcenv.sh
+
+# Prevent MSBuild from messing with PATH, INCLUDE and LIB environment variables.
```
mstorsjo, line 19: Can you extend this sentence a bit, to clarify what "messing with the environment variables" means, so it is understandable even without reading the associated link?

wrappers/msvcenv.sh
```diff
@@ -31,6 +31,7 @@ SDKINCLUDE="$BASE\\$SDK\\include\\$SDKVER"
 SDKLIB="$BASE\\$SDK\\lib\\$SDKVER"
 BINDIR=$BASE_UNIX/vc/tools/msvc/$MSVCVER/bin/Hostx64/$ARCH
 SDKBINDIR=$BASE_UNIX/$SDK_UNIX/bin/$SDKVER/x64
+MSBUILDBINDIR=$BASE_UNIX/MSBuild/Current/Bin/amd64
```
mstorsjo, line 34: Note that while we hardcode the `x64` bindirs above, we overwrite this with a sed command in `install.sh`, see this snippet:
```
host=x64
if [ "$(uname -m)" = "aarch64" ]; then
    host=arm64
fi
cat $ORIG/wrappers/msvcenv.sh | sed 's/MSVCVER=.*/MSVCVER='$MSVCVER/ | sed 's/SDKVER=.*/SDKVER='$SDKVER/ | sed s/x64/$host/ > msvcenv.sh
```
So I'd appreciate if you'd update this accordingly to change the arch from amd64 to arm64 or whatever the corresponding directory is called here as well.

mstorsjo: And as a separate point: For various reasons, we lowercase many other directories, like `vc/tools/msvc/$MSVCVER/bin` above. But if we don't have any concrete reason for doing it here, we can certainly leave it in its original casing, and it's easy to revisit this if we run into issues with mixed casing and case sensitiveness.

test/HelloWorld.vcxproj
```diff
@@ -0,0 +1,106 @@
...
+    </ClCompile>
+    <Link>
+      <GenerateDebugInformation>true</GenerateDebugInformation>
+      <SubSystem>Windows</SubSystem>
+      <OptimizeReferences>true</OptimizeReferences>
+      <EnableCOMDATFolding>true</EnableCOMDATFolding>
+    </Link>
+  </ItemDefinitionGroup>
+  <ItemGroup>
+    <ClCompile Include="HelloWorld.cpp" />
```
mstorsjo, line 101: I'm a little on the fence here, whether it's worth to add another entirely separate test app for this, or if we should reuse the existing `hello.c` here. On the other hand, this doesn't really hurt to have a slightly different testcase, testing C++ and a GUI app instead of a console one. As long as we don't need to try to execute the build output it's probably fine.

ravenexp:
Hmm, I think `test-msbuild.sh` requires `wine-mono` to be installed in the CI test system...

mstorsjo (commented):
Thanks for the PR, overall this looks very neatly integrated. A couple minor inline comments.

mstorsjo:
> Hmm, I think `test-msbuild.sh` requires `wine-mono` to be installed in the CI test system...

Yes, I would think so. I've only briefly looked at msbuild before, and skipped as it was .NET based (and I'm very much not familiar with that whole ecosystem).

For the macos based CI build, it seems to have some suitable version of wine-mono available already (apparently included in whatever we install with homebrew), but on linux we'd probably need to request it separately.

Apart from that, I'd appreciate if `test.sh` could check if mono seems to be available, and skip running `test-msbuild.sh` if not. For the CI, we'd obviously want to actually run it, but for local installs, I'd like to be able to run the tests without bothering with msbuild and mono.

wrappers/msbuild
ravenexp: Rewritten in https://github.com/mstorsjo/msvc-wine/pull/103/commits/bb1b6767cdfda66a2887648cc749960d247b5b9d#diff-af9afe9da8c6e69fc2391bfcb45a4019e23f6d88f0c42b7f870d923e5c9dd44cR19-R22

wrappers/msvcenv.sh
ravenexp: > So I'd appreciate if you'd update this accordingly to change the arch from amd64 to arm64 or whatever the corresponding directory is called here as well.

Done in new commit https://github.com/mstorsjo/msvc-wine/pull/103/commits/5a334b4371e28c943278c5f4466085a5c6c9efa0

ravenexp: > And as a separate point: For various reasons, we lowercase many other directories, like `vc/tools/msvc/$MSVCVER/bin` above. But if we don't have any concrete reason for doing it here, we can certainly leave it in its original casing, and it's easy to revisit this if we run into issues with mixed casing and case sensitiveness.

As for this, I'm not sure. The original casing seems to work in WINE at least as far as `MSBuild.exe` is concerned. And other tools should not touch these directories.

test/HelloWorld.vcxproj
ravenexp: I specifically wanted to test a typical Win32 GUI app that links to many Windows SDK (not VC runtime) libraries like `kernel32`, `user32`, etc. My end goal is to build unmodified VS solutions and projects in the Linux CI using MSBuild.

ravenexp:
> For the macos based CI build, it seems to have some suitable version of wine-mono available already (apparently included in whatever we install with homebrew), but on linux we'd probably need to request it separately.
> 
> Apart from that, I'd appreciate if `test.sh` could check if mono seems to be available, and skip running `test-msbuild.sh` if not. For the CI, we'd obviously want to actually run it, but for local installs, I'd like to be able to run the tests without bothering with msbuild and mono.

It looks like I've accidentally disabled `test-msbuild.sh` for macos as well... Does it not have `/usr/share/wine/` ?

wrappers/msvcenv.sh
mstorsjo: Yeah, this is probably fine - in the long run it's definitely better to alter the case as little as possible, only when specifically needed.

As long as things execute in wine, the case sensitiveness isn't usually an issue. It has been an issue with e.g. the output from `-showIncludes`, if it is used by a case sensitive tool to track dependencies for rebuilds though, but that probably doesn't matter here.

test/HelloWorld.vcxproj
mstorsjo: Yeah, I get that goal. With respect to things that can break here, the choice between a C or C++ app, or console or gui app, doesn't really affect the actual test coverage much (I don't think I can come up with a situation where one of them would work but the other would fail), but for the purpose of making a more typical testcase, I think this is fine, so let's go with this, as it is permissively licensed and all.

mstorsjo:
> > For the macos based CI build, it seems to have some suitable version of wine-mono available already (apparently included in whatever we install with homebrew), but on linux we'd probably need to request it separately.
> > Apart from that, I'd appreciate if `test.sh` could check if mono seems to be available, and skip running `test-msbuild.sh` if not. For the CI, we'd obviously want to actually run it, but for local installs, I'd like to be able to run the tests without bothering with msbuild and mono.
> 
> It looks like I've accidentally disabled `test-msbuild.sh` for macos as well... Does it not have `/usr/share/wine/` ?

On macOS, only the OS itself can touch /usr, package managers like homebrew usually install into `/usr/local` (or in some cases `/opt`).

ravenexp:
I can not get MSBuild tests to work on either of the GitHub runners. They fail with different cryptic errors. Installing 32-bit WINE didn't help.
I'm giving up for now.

Perhaps it can eventually be solved by updating WINE from an ancient v6.0 from 2021 shipped with Ubuntu 22.04 to something more modern.
I have `wine-8.21` and `wine-mono-8.1.0` installed locally and the tests work.
Ubuntu installs `wine-6.0` with `wine-mono-5.1.1` and fails the tests.

mstorsjo:
> Perhaps it can eventually be solved by updating WINE from an ancient v6.0 from 2021 shipped with Ubuntu 22.04 to something more modern. I have `wine-8.21` and `wine-mono-8.1.0` installed locally and the tests work. Ubuntu installs `wine-6.0` with `wine-mono-5.1.1` and fails the tests.

Hmm, that could possibly be the issue, else it's something different in the environment which causes the mono execution to fail.

Is it possible to install a custom version of Wine from WineHQ in the CI runners, just for testing to sort out if this is the root cause?

ravenexp:
> Is it possible to install a custom version of Wine from WineHQ in the CI runners, just for testing to sort out if this is the root cause?

I tried installing stable Wine 8.0 from WineHQ and the MSBuild tests now indeed appear to work on Ubuntu. But then the runner bugged out. Can you perhaps kill and restart it?

ravenexp:
Ok, the GitLab CI tests run finished in 4 minutes. No strange issues. But I do not install ARM MSVC toolchains into my image, so it was only running the first half of the tests.

ravenexp:
I've run Linux tests two more times here on GitHub and there were no issues. MSBuild tests still don't work on macOS, so I didn't enable them in the final commit version. Apart from a seemingly random `test-docker` failure, everything seems to be working now.

From my previous experience, GitHub runners sometimes just hang for no obvious reason, and simply restarting the job fixes that.
I don't think it has anything to do with running Wine.

mstorsjo:
> > Oh, lovely… Issues like that are always a risk when using wine in a CI-like setup. Is the issue reproducible? Does it happen when using the third party wine 8.0 without the whole mono/msbuild additions?
> 
> After 7 more GitHub CI runs, I'm fairly confident that that problem is not reproducible.

Thanks, that's good news.

> Stable WINE 9.0 was released this week, so the CI config required some adjustment and more runs. I've also tried several horrible hacks to get the tests to run on macOS, but it looks like Mono is just hopelessly broken on this system...

Ok, that's totally no problem if we don't run these tests on macOS, as long as we test it successfully somewhere, and get a green CI run elsewhere.

> New commit [5ee766d](https://github.com/mstorsjo/msvc-wine/commit/5ee766db0468b7646e89178d7ce217abc0edcab5) addresses a fairly important issue I initially overlooked when submitting this PR. Today I've finally got kernel driver projects using the WDK to build, so now I'm fairly confident that MSBuild is configured correctly by its wrapper script.
> 
> New commit [303d0bf](https://github.com/mstorsjo/msvc-wine/commit/303d0bfd93a36c6cafbb284c6ddb58f8e0e77489) improves the UX of running MSBuild on Linux. I'm running it a lot now:)

That's good to hear, good to know that it has a bit more trackrecord now!

Sorry I haven't had time to follow up on this in the last few days, I hope to have time to have a closer look again in a few days.

mstorsjo:
Thanks for the update - I think this mostly looks fine! @huangqinjin - are you ok with the last commit of the branch, it touches parts of `wine-msvc.sh` that you've modified?

test/test-msbuild.sh
```diff
@@ -0,0 +1,26 @@
...
+# WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+# ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+# OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+. "${0%/*}/test.sh"
+
+for config in Debug Release; do
+    for useenv in true false; do
+        EXEC "" ${BIN}msbuild /p:UseEnv=${useenv} /p:Configuration=${config} "${TESTS}HelloWorld.vcxproj"
+        rm -rf "${TESTS}/${config}"
```
huangqinjin, line 22: Can we set `IntDir` and `OutDir` to `${CWD}` such that `msbuild` doesn't populate things to source directory?

wrappers/msbuild
```diff
@@ -0,0 +1,70 @@
...
+    ;;
+arm64)
+    export Platform=ARM64
+    ;;
+*)
+    export Platform=${ARCH}
+    ;;
+esac
+
+export WINE_MSVC_RAW_STDOUT=1
```
huangqinjin, line 69: Should we check that current stdout is actually a TTY.

huangqinjin:
@mstorsjo The modification looks good to me.

test/test-msbuild.sh
ravenexp: Good idea! I'll add a new commit addressing this.

wrappers/msbuild
ravenexp: Maybe we could, but is there anything we gain by conditionally enabling post-processing for the MSBuild command output? It is intended for the user, not for the other tools to parse.

mstorsjo: Also, if the output isn’t a tty, the tool itself disables the output coloring on its own (which is the whole reason for this change in the first place) - so perhaps it would be inconsistent if we’d enable filtering off the output only in some cases?

huangqinjin: My first thought was that, msbuild would hang if its stdout and stderr got piped like https://github.com/mstorsjo/msvc-wine/blob/9e9ae8bdb42bedb34ec2415f8f704e43b41762a8/wrappers/wine-msvc.sh#L72

That is why we introduce `msvctricks`. But I think the usecase is rare. So I'm OK with current logic.

test/test-msbuild.sh
ravenexp: Resolved in https://github.com/mstorsjo/msvc-wine/pull/103/commits/345b115b8ee931ec6df457b4cd251412874cb71c.

test/test-msbuild.sh
```diff
@@ -18,8 +18,13 @@
 
 for config in Debug Release; do
     for useenv in true false; do
-        EXEC "" ${BIN}msbuild /p:UseEnv=${useenv} /p:Configuration=${config} "${TESTS}HelloWorld.vcxproj"
-        rm -rf "${TESTS}/${config}"
+        # Trailing slash is required in MSBuild directory properties.
```
mstorsjo, line 21: Nit: there’s a typo in the commit message, “ atrifacts”. LGTM otherwise.

ravenexp: Amended in https://github.com/mstorsjo/msvc-wine/pull/103/commits/1da913e9373f44a717221e6c972a98a737ab5b3d. Thanks!

mstorsjo (approved):
This LGTM now, thanks a bunch!

I’ll merge this later today if there’s no more comments.

PR #104, opened 2024-01-23
test: Enable mt.exe on macOS with Wine 9.0

huangqinjin:
mt.exe no longer crashes on macOS with Wine 9.0.

macOS-12 runner image version 20240119 has updated to Wine 9.0.
https://github.com/huangqinjin/msvc-wine/actions/runs/7583317817/job/20752571482

 Let's wait until most runners have been upgraded.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
     steps:
       - name: Install prerequisites
         run: |
-          brew install wine-stable msitools cmake ninja meson
+          brew update && brew install wine-stable msitools cmake ninja meson
           wine64 wineboot
       - uses: actions/checkout@v3
       - name: Download MSVC
--- a/test/test-cmake.sh
+++ b/test/test-cmake.sh
@@ -30,9 +30,6 @@ CMAKE_ARGS=(
 case $OSTYPE in
     darwin*)
         CMAKE_ARGS+=(
-            # mt.exe always crashes on macOS.
-            -DCMAKE_EXE_LINKER_FLAGS=/MANIFEST:NO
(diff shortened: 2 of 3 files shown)


mstorsjo:
> mt.exe no longer crashes on macOS with Wine 9.0.

That's good news, thanks!

> macOS-12 runner image version 20240119 has updated to Wine 9.0. https://github.com/huangqinjin/msvc-wine/actions/runs/7583317817/job/20752571482
> 
> Let's wait until most runners have been upgraded.

This is weird - as Wine itself isn't shipped as part of the runner install, but we install it from brew, I wouldn't expect that the runner version itself should matter. Or does it depend on how old brew package listing cache there is in the base install? Can we do something like `brew update` or similar (untested) to make sure we get the latest Wine consistently, for now until all active runners have a new enough brew package cache?

huangqinjin:
Yes, `brew update` works!

mstorsjo:
> Yes, `brew update` works!

Thanks, this looks good to me. I testran `brew update && brew install wine-stable` a couple times, so I got a runner with an older OS image, and I still got wine 9.0, so this seems to work as expected.

mstorsjo:
FWIW, the `brew update` stage seems to break installing meson (which installs python) in current versions - see https://github.com/mstorsjo/msvc-wine/actions/runs/7902146137/job/21567372457.

It's reproducible with something as little as `brew update && brew install meson`.

It's not clear to me if this is a generic homebrew issue (I don't see any reports about the same in homebrew), or if it is caused by something else installed on the github actions runners, that conflicts with homebrew's python when it tries to update it. It does seem to happen with both `macos-12` and `macos-13` runners (which both are x86), but not with `macos-14` (which is arm64). AFAIK `macos-latest` currently actually still points at `macos-12`.

Issue #30, opened 2021-10-14
wine64 is needed instead of wine for x64 toolchain on Ubuntu 20.04


dmikushin:
Hi @mstorsjo ,

On Ubuntu 20.04 wine is no longer able to execute 64-bit Windows apps. In order to get it to work, I needed to call wine64 instead of wine explicitly. The change to the script could be like this:

```
Small fix to the Martin's original scripts: wine64 should be used for x64.
RUN sed -i "s/wine /win64 /" /opt/msvc/bin/x64/wine-msvc.sh
```

Thanks！you are so great。(c)

johnsonjh:
Looks like this is fixed and can be closed.

PR #107, opened 2024-01-29
Try using `wine64` then fall back to `wine`

### LINKED ISSUE #99: there seems to be no wine64 in ubuntu23.10
only `wine` not `wine64` in new version ubuntu 23.10
i add a symbol link wine64 -> wine in system

johnsonjh:
Try using `wine64` then fall back `wine`; this should close #99 (and not introduce regressions for cases like #30).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -166,15 +166,17 @@ for arch in x86 x64 arm arm64; do
 done
 rm msvcenv.sh
 
-if [ -d "$DEST/bin/$host" ] && [ -x "$(which wine64 2>/dev/null)" ]; then
-    WINEDEBUG=-all wine64 wineboot &>/dev/null
-    echo "Build msvctricks ..."
-    "$DEST/bin/$host/cl" /EHsc /O2 "$ORIG/msvctricks.cpp"
-    if [ $? -eq 0 ]; then
-        mv msvctricks.exe bin/
-        rm msvctricks.obj
-        echo "Build msvctricks done."
-    else
-        echo "Build msvctricks failed."
+if [ -d "$DEST/bin/$host" ]; then
+    if WINE="$(command -v wine64 || command -v wine)"; then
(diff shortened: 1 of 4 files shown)


install.sh
```diff
@@ -166,8 +166,8 @@ for arch in x86 x64 arm arm64; do
 done
 rm msvcenv.sh
 
-if [ -d "$DEST/bin/$host" ] && [ -x "$(which wine64 2>/dev/null)" ]; then
-    WINEDEBUG=-all wine64 wineboot &>/dev/null
+if [ -d "$DEST/bin/$host" ]; then
+    WINEDEBUG=-all $(command -v wine64 || command -v wine || false) wineboot &>/dev/null
```
mstorsjo, line 170: This change isn't right; if there's no wine installed, this will end up trying to still execute something and fail.

This breaks in the clang/lld testcase in CI, where we explicitly don't install wine and don't want to use that, we're just going to use the MSVC headers/libraries with clang/lld.

mstorsjo (commented):
This looks mostly ok, but there's one change which breaks one test in the CI.

Also, I'd prefer if you'd move the "closes #99" bit from the commit message, from the subject line into a separate line in the body of the commit message.

install.sh
johnsonjh: Ah, I was executing "false" purposely, so if wine64 or wine wasn't installed, it would run "false" to explicitly cause an error here, assuming that not having wine installed should be fatal.  I'll fix this up.

johnsonjh: I am assuming this is the only place this would be needed - if not let me know.

I also didn't change the README file - I don't think you intended to be that pedantic since that's human readable docs.  I'll make these changes now.

mstorsjo (approved):
LGTM, thanks!

PR #109, opened 2024-02-02
msbuild: Prefer x86_64 build tools for all targets

### LINKED ISSUE #108: msbuild uses HostX86 tools when cross-compiling on x64 host and UseEnv=false
x86: https://github.com/mstorsjo/msvc-wine/actions/runs/7688417957/job/20949618574#step:5:169
>   z:\home\runner\work\msvc-wine\msvc-wine\msvc\vc\tools\msvc\14.38.33130\bin\HostX86\x86\CL.exe /c /Zi /nologo /W3 /WX- /diagnostics:column /Od /Oy- /D WIN32 /D _DEBUG /D _WINDOWS /D _UNICODE /D UNICODE /Gm- /EHsc /RTC1 /MDd /GS /fp:precise /Zc:wchar_t /Zc:forScope /Zc:inline /Fo"Z:/tmp/msvc-wine.tmp.BKU5/useenv-false/Debug/" /Fd"Z:/tmp/msvc-wine.tmp.BKU5/useenv-false/Debug/vc143.pdb" /external:W3 /Gd /TP /analyze- /FC /errorReport:queue HelloWorld.cpp

arm: https://github.com/mstorsjo/msvc-wine/actions/runs/7688417957/job/20949618574#step:5:639
>   z:\home\runner\work\msvc-wine\msvc-wine\msvc\vc\tools\msvc\14.38.33130\bin\HostX86\arm\CL.exe /c /Zi /nologo /W3 /WX- /diagnostics:column /Od /Oy- /D WIN32 /D _DEBUG /D _WINDOWS /D _ARM_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE=1 /D _UNICODE /D UNICODE /Gm- /EHsc /RTC1 /MDd /GS /fp:precise /Zc:wchar_t /Zc:forScope /Zc:inline /Fo"Z:/tmp/msvc-wine.tmp.nS3F/useenv-false/Debug/" /Fd"Z:/tmp/msvc-wine.tmp.nS3F/useenv-false/Debug/vc143.pdb" /external:W3 /Gd /TP /analyze- /FC /errorReport:queue HelloWorld.cpp

arm64: https://github.com/mstorsjo/msvc-wine/actions/runs/7688417957/job/20949618574#step:5:872
>   z:\home\runner\work\msvc-wine\msvc-wine\msvc\vc\tools\msvc\14.38.33130\bin\HostX86\arm64\CL.exe /c /Zi /nologo /W3 /WX- /diagnostics:column /Od /Oy- /D WIN32 /D _DEBUG /D _WINDOWS /D _ARM64_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE=1 /D _UNICODE /D UNICODE /Gm- /EHsc /RTC1 /MDd /GS /fp:precise /Zc:wchar_t /Zc:forScope /Zc:inline /Fo"Z:/tmp/msvc-wine.tmp.8ncV/useenv-false/Debug/" /Fd"Z:/tmp/msvc-wine.tmp.8ncV/useenv-false/Debug/vc143.pdb" /external:W3 /Gd /TP /analyze- /FC /errorReport:queue HelloWorld.cpp

CC @ravenexp

ravenexp:
Set a global MSBuild property to always prefer 64-bit build tools.

Reference:
https://learn.microsoft.com/en-us/cpp/build/reference/msbuild-visual-cpp-overview?view=msvc-170#preferredtoolarchitecture-property

Resolves #108

### DIFF
--- a/wrappers/msbuild
+++ b/wrappers/msbuild
@@ -66,5 +66,10 @@ arm64)
     ;;
 esac
 
+if [[ "${MSBUILDBINDIR}" == *amd64 ]]; then
+    # Prefer x86_64 build tools for any target arch.
+    export PreferredToolArchitecture=x64
+fi
+
 export WINE_MSVC_RAW_STDOUT=1
 $(dirname $0)/wine-msvc.sh ${MSBUILDBINDIR}/MSBuild.exe "$@"


wrappers/msbuild
```diff
@@ -66,5 +66,8 @@ arm64)
     ;;
 esac
 
+# Prefer x86_64 build tools for any target arch.
+export PreferredToolArchitecture=x64
```
huangqinjin, line 70: Is it working on arm64 linux @mstorsjo

ravenexp: Is there an arm64 port of Wine-Mono though? MSBuild is a .NET 4.x executable, so it needs Mono to run on Wine.

There are no arm64 Mono releases, at least here: https://github.com/madewokherd/wine-mono/releases/tag/wine-mono-9.0.0

mstorsjo: I haven’t tested anything with mono/msbuild yet, no. While there might not be any release of it, I presume it could be possible to build it (but I haven’t spent effort in trying to set that up).

So it think we should have the same logic here, as we have in install.sh, for picking the host executable arch - even if actually testing it on arm64 still is up in the air. (Keeping it pointing at x64 won’t work in any case, while pointing at arm64 probably should work out and be correct, even if there may be other challenges around mono on arm64.)

ravenexp: > Keeping it pointing at x64 won’t work in any case, while pointing at arm64 probably should work out and be correct, even if there may be other challenges around mono on arm64.

I studied MSBuild VC property files some more, and it looks like arm64 tools are always preferred on arm64 hosts, so there is no need to set this property in that case.

I'll wrap it in
```sh
if [[ `uname -m` == "x86_64" ]];
```
just to be safe.

mstorsjo (approved):
LGTM, thanks!

wrappers/msbuild
```diff
@@ -66,5 +66,10 @@ arm64)
     ;;
 esac
 
+if [[ "$(uname -m)" == "x86_64" ]]; then
```
huangqinjin, line 69: Is it better to test `"$BINDIR" contains "Hostx64"`?

ravenexp: I've added a test for `$MSBUILDBINDIR` because it's easier to match.

huangqinjin: LGTM!

Issue #28, opened 2021-10-08
Kernel-mode headers


maharmstone:
Thanks for this, very cool!

Do you know of any way to get the kernel-mode headers as well? There should be a `km` directory, alongside `um` and `shared`, containing `ntifs.h` and the like (i.e. what's in /usr/include/ddk for mingw). I had a brief look in Microsoft's package listing but couldn't see anything obvious.

maharmstone:
Thanks @Biswa96, that was helpful. The files are in `Installers/Windows Driver Kit Headers and Libs-x86_en-us.msi` and `Installers/Windows Driver Kit ARM Headers and Libs-x86_en-us.msi` of the ISO.

maharmstone:
Thanks Martin and @ravenexp for this - I'll give it a go. It should make my workflow a lot simpler.

ravenexp:
> Thanks Martin and @ravenexp for this - I'll give it a go. It should make my workflow a lot simpler.

The main remaining blocking issue for me is that I can not sign the drivers I build on Linux. Both `inf2cat.exe` and `signtool.exe` do not work in WINE yet.

There are OSS alternatives such as https://github.com/LINBIT/generate-cat-file and https://github.com/mtrojnar/osslsigncode, but I still haven't got around to try them. If you have any success with these tools, then we could maybe add links to them to the README.

maharmstone:
> The main remaining blocking issue for me is that I can not sign the drivers I build on Linux. Both `inf2cat.exe` and `signtool.exe` do not work in WINE yet.
> 
> There are OSS alternatives such as https://github.com/LINBIT/generate-cat-file and https://github.com/mtrojnar/osslsigncode, but I still haven't got around to try them. If you have any success with these tools, then we could maybe add links to them to the README.

Yes, this was something that was always problematic for me in the past. I could never get osslsigncode to work. I'd not heard of generate-cat-file, but it looks like it only does SHA1 hashes, and IIRC some versions of Windows require SHA256.

maharmstone:
I couldn't find any free way to make cat files, so ended up making my own clone of Microsoft's makecat: https://github.com/maharmstone/nyan.

ravenexp:
> I couldn't find any free way to make cat files, so ended up making my own clone of Microsoft's makecat: https://github.com/maharmstone/nyan.

[...]

PR #111, opened 2024-02-05
RFC: Add basic support for automated WDK installation

ravenexp:
Add a new command line option '--with-wdk-installers' pointing to the directory where the MSI packages downloaded by `wdksetup.exe` are located.

Extract the WDK packages after unpacking the Windows SDK packages.
They both install files in the same versioned directories.

Unpack the included VS extension files onto the MSVC tree.

Fix the WDK directory structure after extracting MSI packages with `msiextract` on Linux.

Resolves https://github.com/mstorsjo/msvc-wine/issues/28

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -125,17 +125,32 @@ echo Using SDK version $SDKVER
 # The original casing of file names is preserved though, by adding lowercase
 # symlinks instead of doing a plain rename, so files can be referred to with
 # either the out of the box filename or with the lowercase name.
-$ORIG/lowercase -map_winsdk -symlink kits/10/include/$SDKVER/um
-$ORIG/lowercase -map_winsdk -symlink kits/10/include/$SDKVER/shared
-$ORIG/lowercase -map_winsdk -symlink kits/10/include/$SDKVER/winrt
-$ORIG/fixinclude -map_winsdk kits/10/include/$SDKVER/um
-$ORIG/fixinclude -map_winsdk kits/10/include/$SDKVER/shared
-$ORIG/fixinclude -map_winsdk kits/10/include/$SDKVER/winrt
-for arch in x86 x64 arm arm64; do
-    if [ ! -d "kits/10/lib/$SDKVER/um/$arch" ]; then
-        continue
+for incdir in um shared winrt km; do
+    SDK_INCDIR="kits/10/include/$SDKVER/$incdir"
+
+    if [ -d "$SDK_INCDIR" ]; then
+        $ORIG/lowercase -map_winsdk -symlink "$SDK_INCDIR"
+        $ORIG/fixinclude -map_winsdk "$SDK_INCDIR"
     fi
-    $ORIG/lowercase -symlink kits/10/lib/$SDKVER/um/$arch
+done
+
+# The WDF is a part of the Windows Driver Kit.
+WDF_INCDIR="kits/10/include/wdf"
+if [ -d "$WDF_INCDIR" ]; then
+    $ORIG/lowercase -map_winsdk -symlink "$WDF_INCDIR"
+    $ORIG/fixinclude -map_winsdk "$WDF_INCDIR"
+fi
+
+for arch in x86 x64 arm arm64; do
+    SDK_LIBDIR="kits/10/lib/$SDKVER/um/$arch"
+    DDK_LIBDIR="kits/10/lib/$SDKVER/km/$arch"
+
+    if [ -d "$SDK_LIBDIR" ]; then
+        $ORIG/lowercase -symlink "$SDK_LIBDIR"
+    fi
+    if [ -d "$DDK_LIBDIR" ]; then
+        $ORIG/lowercase -symlink "$DDK_LIBDIR"
+    fi
 done
 
 host=x64
--- a/test/test.sh
+++ b/test/test.sh
@@ -109,6 +109,10 @@ for arch in x86 x64 arm arm64; do
          || -d /usr/local/share/wine/mono \
          || -d /opt/wine/mono ]]; then
         EXEC "" BIN=$BIN ./test-msbuild.sh
+
+        if [[ -n $HAVE_WDK && ($arch == "x64" || $arch == "arm64") ]]; then
+            EXEC "" BIN=$BIN ./test-wdk-msbuild.sh
+        fi
     fi
 done
 
--- a/wrappers/msvcenv.sh
+++ b/wrappers/msvcenv.sh
@@ -34,8 +34,8 @@ SDKLIB="$SDKBASE\\lib\\$SDKVER"
 BINDIR=$BASE_UNIX/vc/tools/msvc/$MSVCVER/bin/Hostx64/$ARCH
 SDKBINDIR=$BASE_UNIX/$SDK_UNIX/bin/$SDKVER/x64
 MSBUILDBINDIR=$BASE_UNIX/MSBuild/Current/Bin/amd64
-export INCLUDE="$MSVCDIR\\include;$SDKINCLUDE\\shared;$SDKINCLUDE\\ucrt;$SDKINCLUDE\\um;$SDKINCLUDE\\winrt"
-export LIB="$MSVCDIR\\lib\\$ARCH;$SDKLIB\\ucrt\\$ARCH;$SDKLIB\\um\\$ARCH"
+export INCLUDE="$MSVCDIR\\include;$SDKINCLUDE\\shared;$SDKINCLUDE\\ucrt;$SDKINCLUDE\\um;$SDKINCLUDE\\winrt;$SDKINCLUDE\\km"
+export LIB="$MSVCDIR\\lib\\$ARCH;$SDKLIB\\ucrt\\$ARCH;$SDKLIB\\um\\$ARCH;$SDKLIB\\km\\$ARCH"
 export LIBPATH="$LIB"
 # "$MSVCDIR\\bin\\Hostx64\\x64" is included in PATH for DLLs.
 export WINEPATH="${BINDIR//\//\\};${SDKBINDIR//\//\\};$MSVCDIR\\bin\\Hostx64\\x64"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,6 +3,10 @@ on:
   push:
   pull_request:
 
+env:
+  # From https://learn.microsoft.com/en-us/windows-hardware/drivers/download-the-wdk#download-icon-step-3-install-wdk
+  WDK_INSTALLER_URL: "https://go.microsoft.com/fwlink/?linkid=2249371"
+
 jobs:
   test-msvc-wine-linux:
     runs-on: ubuntu-latest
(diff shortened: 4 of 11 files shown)


install.sh
```diff
@@ -125,17 +125,39 @@ echo Using SDK version $SDKVER
...
+    if [ -d "$SDK_INCDIR" ]; then
+        $ORIG/lowercase -map_winsdk -symlink "$SDK_INCDIR"
+    fi
+done
+
+for incdir in um shared winrt km; do
+    SDK_INCDIR="kits/10/include/$SDKVER/$incdir"
+
+    if [ -d "$SDK_INCDIR" ]; then
+        $ORIG/fixinclude -map_winsdk "$SDK_INCDIR"
```
mstorsjo, line 140: Can you merge these two loops? It would deviate from the order in the script originally, but I think it should be harmless (fixinclude doesn't look at files outside of the given directory).

wrappers/msvcenv.sh
```diff
@@ -34,8 +34,8 @@ SDKLIB="$SDKBASE\\lib\\$SDKVER"
 BINDIR=$BASE_UNIX/vc/tools/msvc/$MSVCVER/bin/Hostx64/$ARCH
 SDKBINDIR=$BASE_UNIX/$SDK_UNIX/bin/$SDKVER/x64
 MSBUILDBINDIR=$BASE_UNIX/MSBuild/Current/Bin/amd64
-export INCLUDE="$MSVCDIR\\include;$SDKINCLUDE\\shared;$SDKINCLUDE\\ucrt;$SDKINCLUDE\\um;$SDKINCLUDE\\winrt"
-export LIB="$MSVCDIR\\lib\\$ARCH;$SDKLIB\\ucrt\\$ARCH;$SDKLIB\\um\\$ARCH"
+export INCLUDE="$MSVCDIR\\include;$SDKINCLUDE\\shared;$SDKINCLUDE\\ucrt;$SDKINCLUDE\\um;$SDKINCLUDE\\winrt;$SDKINCLUDE\\km"
+export LIB="$MSVCDIR\\lib\\$ARCH;$SDKLIB\\ucrt\\$ARCH;$SDKLIB\\um\\$ARCH;$SDKLIB\\km\\$ARCH"
```
mstorsjo, line 38: Curious question; if installing the WDK on Windows, does one get both `um` and `km` directories added to `$INCLUDE` and `$LIB` there as well, in e.g. a developer command prompt?

mstorsjo (commented):
Overall, this looks very good, thanks!

Could we add some minimal testing of it in the CI setup? E.g. enable the `--with-wdk-installers` in one setup, and then e.g. try compiling and linking some minimal test driver? (Doesn't need to be a full test driver from somewhere, just something minimal with a single entry point function or similar, would be enough, as long as it testruns including a kernel mode header and linking a kernel mode library.)

As you're adding support for it in msbuild, I presume that's your actual use case, so it might be good with a test for that as well (although I'm not sure if I'd consider it mandatory here), but it would be good with a separate plain test of just one straight `cl.exe` invocation as well, to test whether files were installed correctly.

Not sure what the best way is to detect whether this actually was meant to be installed though. Maybe a flag like `HAVE_WDK=1 ./test.sh`? Since we do add paths like `km` to the INCLUDE path unconditionally, even if they weren't installed, we can't check if they're present in the INCLUDE variable. And if we check whether including such a header succeeds, the test just remains silent if we fail to install what we meant to.

install.sh
ravenexp: I was not sure whether there are dependencies between `lowercase` and `fixinclude`, and just assumed that all headers must be lowercased for `fixinclude` to work. I'll merge them into one loop, then.

wrappers/msvcenv.sh
ravenexp: I actually have not checked that, because I've only built drivers via MSBuild, and it does not use these variables. Unfortunately, msvc-wine does not install `vsdevcmd.bat`, so I can not check it here on Linux. I'll look into it when I get to a Windows machine.

ravenexp:
> Could we add some minimal testing of it in the CI setup? E.g. enable the --with-wdk-installers in one setup, and then e.g. try compiling and linking some minimal test driver? (Doesn't need to be a full test driver from somewhere, just something minimal with a single entry point function or similar, would be enough, as long as it testruns including a kernel mode header and linking a kernel mode library.)

I thought about doing this, but it will add ~700 MB of downloads to the CI job and ~ 1.5 GB to the install size.
It also requires downloading `wdksetup.exe` from the Microsoft site and running it in automated mode in WINE.
Initially I had all this implemented directly in `vsdownload.py`, but then decided against it and changed it to work with just the `.msi` packages that are _somehow_ obtained. Perhaps I could add a small helper script that downloads `wdksetup.exe`, runs it with the right flags and downloads the WDK packages directly into the cache directory?

As for the test driver, I could take a MIT licensed example driver project from Microsoft GitHub, just as I did with MSBuild tests.
They only provide `.vcxproj` build files, but I can add a `Makefile` or `CMakeLists.txt` to do the same thing with plain `cl.exe` and `link.exe`.

mstorsjo:
> I thought about doing this, but it will add ~700 MB of downloads to the CI job and ~ 1.5 GB to the install size.

Right, it probably would make the CI job a bit slower. The install size should probably be fine, as we don't distribute the installed directory anywhere anyway (and aren't allowed to anyway). OTOH if it needlessly hammers Microsoft servers, where they aren't prepared to handle very frequent extra downloads, it might not be so nice...

> It also requires downloading `wdksetup.exe` from the Microsoft site and running it in automated mode in WINE. Initially I had all this implemented directly in `vsdownload.py`, but then decided against it and changed it to work with just the `.msi` packages that are _somehow_ obtained.

Oh, I missed this detail.

> Perhaps I could add a small helper script that downloads `wdksetup.exe`, runs it with the right flags and downloads the WDK packages directly into the cache directory?

That would certainly be appreciated indeed.

> As for the test driver, I could take a MIT licensed example driver project from Microsoft GitHub, just as I did with MSBuild tests. They only provide `.vcxproj` build files, but I can add a `Makefile` or `CMakeLists.txt` to do the same thing with plain `cl.exe` and `link.exe`.

Sure, that would probably be good.

ravenexp:
I getting strange `mergeTrees()` failures on macOS. I don't know where to start debugging this, since this command is run by the native Python, not in WINE. Is it OK to skip testing WDK install on this platform?

mstorsjo:
> I getting strange `mergeTrees()` failures on macOS. I don't know where to start debugging this, since this command is run by the native Python, not in WINE. Is it OK to skip testing WDK install on this platform?

That sounds odd.

Yes, it’s ok to not test it on all platforms, as long as it’s tested in at least one configuration in CI.

That said, if you want to debug it, I would add lots of debug printouts in vsdownload.py, to show what files it sees and how it progrfilesesses. But it’s a bit surprising to hear about such failures in unpacking(?) files. I wonder if it’s due to a case insensitive file system, or if it is due to a different version of Python.

ravenexp:
> But it’s a bit surprising to hear about such failures in unpacking(?) files. I wonder if it’s due to a case insensitive file system, or if it is due to a different version of Python.

Wait, does macOS have a case-insensitive file system? Isn't it a kind of Unix system since Mac OS X days?
That would definitely explain it, because the failing code merges "Build" and "build" sub-directories created in the same directory.

mstorsjo:
> > But it’s a bit surprising to hear about such failures in unpacking(?) files. I wonder if it’s due to a case insensitive file system, or if it is due to a different version of Python.
> 
> Wait, does macOS have a case-insensitive file system? Isn't it a kind of Unix system since Mac OS X days? That would definitely explain it, because the failing code merges "Build" and "build" sub-directories created in the same directory.

Yes, it is very much a Unix (even a certified one), but they have retained the case insensitiveness. It is possible to set the apple file systems to case sensitive mode too (although I think it might be kinda cumbersome as I’m not sure if it’s possible to change it other than when formatting a partition), but I’ve heard issues with e.g. some Adobe applications misbehaving on case sensitive file systems. For external volumes it’s certainly doable though.

ravenexp:
This is very unfortunate, because my `/build/` tree merging logic pretty much depends on being executed on a case-sensitive file system, since the extracted `Build` and `build` directories must be treated differently due to some bug in the WDK MSI installers.

wrappers/msvcenv.sh
ravenexp: I've run `vcvars64.bat` from a VC install where the WDK was also installed, and the `\km\` paths were not added to `INCLUDE` and `LIB`. Now I'm not so sure that adding them to default paths was the right thing to do. It could simplify writing plain Makefiles and CMakeLists.txt for the drivers though, without resorting to something like https://github.com/SergiusTheBest/FindWDK.

ravenexp:
I've rebased on master and fixed the Ubuntu CI job. macOS tests are not enabled because the WDK installation does not work on case-insensitive file systems with the current implementation. Is it OK to leave this as a Linux-only feature? I don't know if there's any demand for it, since developing Windows drivers on macOS is not something people usually do.

mstorsjo:
> I've rebased on master and fixed the Ubuntu CI job. macOS tests are not enabled because the WDK installation does not work on case-insensitive file systems with the current implementation. Is it OK to leave this as a Linux-only feature? I don't know if there's any demand for it, since developing Windows drivers on macOS is not something people usually do.

Yes, that's totally ok - if it's a less central feature, I'm quite happy with it being tested only in one configuration. As long as it is tested in at least one configuration, it won't be dead code that I don't know how to test.

I'll try to get to re-reviewing this PR within a few days, hopefully!

test/test.sh
```diff
@@ -109,6 +109,10 @@ for arch in x86 x64 arm arm64; do
          || -d /usr/local/share/wine/mono \
          || -d /opt/wine/mono ]]; then
         EXEC "" BIN=$BIN ./test-msbuild.sh
+
+        if [[ -n $HAVE_WDK && ($arch == "x64" || $arch == "arm64") ]]; then
```
mstorsjo, line 113: I presume this limitation on arch is based on what is supported in the vcxproj file?

Is it possible to do a minimal-ish command line compilation of the same as well? (Not required as part of this PR, but would be nice to have.)

wrappers/msvcenv.sh
mstorsjo: To follow up on this; I think it's probably fine to add them by default here. The majority of users won't install it anyway, so it won't bother them. But for those who do install the WDK, I guess there aren't any conflicts between the sets of headers, so that users would need to worry about which set of headers is included first? If there are no conflicts, this is probably best.

(Otherwise, it'd be really messy for user projects to try to figure out what paths to add, when running within msvc-wine.)

mstorsjo (approved):
Looking really good now, thanks! Just a couple discussion points left, otherwise ok!

test/test.sh
ravenexp: I think `x86_64` and `aarch64` are the only architectures supported by current Windows kernels. There used to be 32-bit versions of Windows 10, but it's not something people care about in 2024.

As for the command line test, I wanted to add a plain Makefile, but it turned out it requires an enormous amount of compile and link flags even for the simplest drivers. And many of them are WDK version specific, as you can see here: https://github.com/mstorsjo/msvc-wine/actions/runs/7986251670/job/21806269948#step:5:568 Such a test would probably be very fragile wrt. WDK updates.

wrappers/msvcenv.sh
ravenexp: As far as I know there are no name conflicts between the `km` and `um` headers and libraries. There's a separate set of standard C library headers for the kernel installed in `km/crt`, but this directory is not added to the default include path.

mstorsjo: Ok, thanks! I guess that concludes my questions then - I'll go ahead and merge it then!

test/test.sh
maharmstone: @ravenexp, all versions of Windows 10 support 32-bit architectures. It's only Windows 11 that doesn't.

ravenexp: Oh, I didn't know that. I can add `x86` back to the test matrix if your use case requires this target.

PR #115, opened 2024-02-18
Revert "github: Run brew update to get latest softwares"

huangqinjin:
This reverts commit 26d574b051ddf4642cf92fb92e921c0561196b8a.

Running `brew update` breaks macos-12 runner version 20240202.
It was needed at that moment to get latest wine 9.0 but now it is no longer needed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
     steps:
       - name: Install prerequisites
         run: |
-          brew update && brew install wine-stable msitools cmake ninja meson
+          brew install wine-stable msitools cmake ninja meson
           wine64 wineboot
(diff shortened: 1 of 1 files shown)


mstorsjo:
LGTM, thanks.

Indeed, this was only meant to be a temporary measure.

I tested installing `wine-stable` without `brew update` on the macOS runners, and on `macos-12` (which still, curiously, is the one chosen by `macos-latest`) and `macos-13`, we get runner images dated 20240202 or newer, which all give wine 9.0 (in a test round with 3 attempts). On `macos-14` runners (which we don't use currently), we still get 20240116 images, which give wine 8.0.1 though, but that's probably ok for now.

The brew install issues with python dependencies might not entirely go away (as on `macos-13` runners, just doing `brew install meson`, without any update, fails due to the python dependency at the moment, while it's fine on `macos-12`), but that's not our fault anyway, and should be resolved by some part at some point anyway.

PR #117, opened 2024-03-01
Drop unmaintained Python 2 support via six

ravenexp:
Python 3 style print functions are already used throughout the codebase, and only a few places use six.
Python 2 compatibility is not longer there.

Use the standard `json` module in place of `simplejson`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Install prerequisites
         run: |
           sudo dpkg --add-architecture i386 && sudo apt-get update
-          sudo apt-get install wine64 wine32 python3 msitools python3-simplejson python3-six ca-certificates cmake ninja-build winbind meson
+          sudo apt-get install wine64 wine32 python3 msitools ca-certificates cmake ninja-build winbind meson
           wine64 wineboot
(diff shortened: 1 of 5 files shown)


mstorsjo (approved):
Generally, LGTM, the python2 support isn't tested anywhere any longer so we don't need to keep it.

Out of curiosity - you say it is broken - is this observed empirically, or just assumed? I did test and develop this script in both python2 and python3 form initially, and as far as I remember, there hasn't been all that many bigger changes on top it since that. It'd be interesting to know at which point it did break. (Not that python2->3 transitional knowledge is all that valuable any longer...)

ravenexp:
> Out of curiosity - you say it is broken - is this observed empirically, or just assumed? I did test and develop this script in both python2 and python3 form initially, and as far as I remember, there hasn't been all that many bigger changes on top it since that. It'd be interesting to know at which point it did break. (Not that python2->3 transitional knowledge is all that valuable any longer...)

I have not tested it with `python2` because that needs `python2-six` and `python2-simplejson` packages my distro no longer provides. My assumption was based on the observation that both `six.print_("")` and `print()` functions are being used, but only the former works in Python 2, since `print` is a special statement in v2.

ravenexp:
Oh, you are right! I have not written any Python 2 code myself, so it was all based on what I've read somewhere. I'll replace "broken" -> "unmaintained" everywhere.

PR #96, opened 2023-11-27
Escape absolute paths with Z: also in response files

marcoesposito1988:
Solves the issue discussed in #93

PR #130, opened 2024-06-10
Support manifest filepaths

nolange:
This will add support for mapping paths with the [MANIFESTINPUT Argument](https://learn.microsoft.com/en-us/cpp/build/reference/manifestinput-specify-manifest-input).

### DIFF
--- a/wrappers/wine-msvc.sh
+++ b/wrappers/wine-msvc.sh
@@ -15,49 +15,42 @@
 # OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
 MSVCTRICKS_EXE="$(dirname $0)/../msvctricks.exe"
-EXE="$1"
+EXE=$1
 shift
 
 ARGS=()
-while [ $# -gt 0 ]; do
-	a=$1
-	case $a in
+for a; do
+	path=
+	case "$a" in
 	[-/][A-Za-z]/*)
-		opt=${a:0:2}
-		path=${a:2}
+		path=${a#??}
 		# Rewrite options like -I/absolute/path into -Iz:/absolute/path.
 		# This is needed to avoid what seems like a cl.exe/Wine bug combination
 		# in some very rare cases, see https://bugs.winehq.org/show_bug.cgi?id=55200
 		# for details. In those rare cases, cl.exe fails to find includes in
 		# some directories specified with -I/absolute/path but does find them if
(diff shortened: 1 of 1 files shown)


mstorsjo:
This looks reasonable to me - any comments from @huangqinjin?

wrappers/wine-msvc.sh
```diff
@@ -15,49 +15,46 @@
...
-while [ $# -gt 0 ]; do
-	a=$1
-	case $a in
+for a; do
+	path=
+	case "$a" in
 	[-/][A-Za-z]/*)
-		opt=${a:0:2}
-		path=${a:2}
+		path=${a#??}
```
huangqinjin, line 26: any benefit using this form instead of the original one that substring by pos:len ?

huangqinjin (commented):
It is acceptable to me. But can we make it better? Firstly, the linker options are case-insensitive. Secondly all linker options are in form of `/OPTION:arg`, we can take this opportunity to handle all linker options.

huangqinjin:
We already have the idea in https://github.com/mstorsjo/msvc-wine/pull/96#issuecomment-1858684632

wrappers/wine-msvc.sh
nolange: Broader compatibility with shells (POSIX standard), otherwise no

nolange:
> It is acceptable to me. But can we make it better? Firstly, the linker options are case-insensitive. Secondly all linker options are in form of `/OPTION:arg`, we can take this opportunity to handle all linker options.

Do we want to handle **all linker options**, by guessing if the argument is a filename? could cause more trouble than what it is worth.

huangqinjin:
> Do we want to handle **all linker options**, by guessing if the argument is a filename? could cause more trouble than what it is worth.

I think we are already doing so. @mstorsjo can you comment on this?

mstorsjo:
> > Do we want to handle **all linker options**, by guessing if the argument is a filename? could cause more trouble than what it is worth.
> 
> I think we are already doing so. @mstorsjo can you comment on this?

Yeah, we already do kind of check whether the argument value seems like a pathname, so I think it might be ok to extend it to a pattern rule like this.

@nolange The updated patch looks good, but you can split it back into two commits - one for factorizing the existing code, and one for adding the new case?

mstorsjo:
Thanks, this looks good to me. I'll merge it later if it passes all the CI checks, unless there's any further opinions on it.

PR #136, opened 2024-09-11
vsdownload: kits -> Windows Kits

huangqinjin:
So that the dest dir is valid for clang /winsysroot without installation on Unix and Windows.


mstorsjo:
This looks reasonable, thanks!

PR #140, opened 2024-09-16
vsdownload: Handle machineArch dependency constraint

### LINKED ISSUE #135: wrong arch of some packages
For example, the C# compiler in `MSBuild/Current/Bin/Roslyn/csc.exe` is a `arm64` executable currently. From the manifest file and listing files, it is in package `Microsoft.CodeAnalysis.Compilers` which is a dependency of `Microsoft.VisualStudio.Component.Roslyn.Compiler` 
```json
			"id": "Microsoft.VisualStudio.Component.Roslyn.Compiler",
			"version": "17.11.35102.94",
			"type": "Component",
			"dependencies": {
				"Microsoft.CodeAnalysis.Compilers": "[4.0,5.0.0.0)"
			},
```
and the arch of the package is specified by `productArch`:
```json
			"id": "Microsoft.CodeAnalysis.Compilers",
			"version": "4.11.0.2442804",
			"type": "Vsix",
			"productArch": "arm64",
```

So, I believe we should match `productArch` with host arch. Besides, there are `chip` and `machineArch` fields in the manifest file, and I failed to understand what are the exact meaning of them, e.g. 
```json
			"id": "Microsoft.VisualStudio.ClientDiagnostics.Targeted",
			"version": "17.11.35102.94",
			"type": "Vsix",
			"chip": "x64",
			"productArch": "neutral",
			"machineArch": "x64",
```

### LINKED ISSUE #138: RecursionError: maximum recursion depth exceeded while calling a Python object
Hi @mstorsjo 

Running this command gives error:

`python vsdownload.py --accept-license --print-deps-tree Microsoft.VisualStudio.Workload.VCTools`

```
Microsoft.VisualStudio.PackageGroup.Debugger.Core
        Microsoft.VisualStudio.Debugger
          Microsoft.VisualStudio.Debugger.Remote.DiagnosticsHub.Client
            Microsoft.VisualStudio.Debugger.Remote.DiagnosticsHub.Client
              Microsoft.VisualStudio.Debugger.Remote.DiagnosticsHub.Client
                Microsoft.VisualStudio.Debugger.Remote.DiagnosticsHub.Client
                  Microsoft.VisualStudio.Debugger.Remote.DiagnosticsHub.Client
                    Microsoft.VisualStudio.Debugger.Remote.DiagnosticsHub.Client
                      Microsoft.VisualStudio.Debugger.Remote.DiagnosticsHub.Client
```

```
Traceback (most recent call last):
  File "/tmp/msvc-wine/vsdownload.py", line 654, in <module>
    printDepends(packages, i, "", None, "", args)
  File "/tmp/msvc-wine/vsdownload.py", line 295, in printDepends
    printDepends(packages, key, type, chip, indent + "  ", args)
  File "/tmp/msvc-wine/vsdownload.py", line 295, in printDepends
    printDepends(packages, key, type, chip, indent + "  ", args)
  File "/tmp/msvc-wine/vsdownload.py", line 295, in printDepends
    printDepends(packages, key, type, chip, indent + "  ", args)
  [Previous line repeated 994 more times]
  File "/tmp/msvc-wine/vsdownload.py", line 282, in printDepends
    p = findPackage(packages, target, chip)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/msvc-wine/vsdownload.py", line 249, in findPackage
    id = id.lower()
         ^^^^^^^^^^
RecursionError: maximum recursion depth exceeded while calling a Python object
```

huangqinjin:
Closes #135 
Closes #138

### DIFF
--- a/vsdownload.py
+++ b/vsdownload.py
@@ -21,6 +21,7 @@ import hashlib
 import os
 import multiprocessing.pool
 import json
+import platform
 import shutil
 import socket
 import subprocess
@@ -55,6 +56,7 @@ def getArgsParser():
     parser.add_argument("--msvc-version", metavar="version", help="Install a specific MSVC toolchain version")
     parser.add_argument("--sdk-version", metavar="version", help="Install a specific Windows SDK version")
     parser.add_argument("--with-wdk-installers", metavar="dir", help="Install Windows Driver Kit using the provided MSI installers")
+    parser.add_argument("--host-arch", metavar="arch", choices=["x86", "x64", "arm64"], help="Specify the host architecture of packages to install")
     return parser
 
 def setPackageSelectionMSVC16(args, packages, userversion, sdk, toolversion, defaultPackages):
@@ -217,14 +219,22 @@ def getManifest(args):
 
     return manifest
 
-def prioritizePackage(a, b):
-    if "chip" in a and "chip" in b:
-        ax64 = a["chip"].lower() == "x64"
-        bx64 = b["chip"].lower() == "x64"
-        if ax64 and not bx64:
+def prioritizePackage(arch, a, b):
+    def archOrd(k, x):
+        if arch is None:
+            return 0
+        c = x.get(k, "neutral").lower()
+        if c == "neutral":
+            return 0
+        if c == arch:
             return -1
-        elif bx64 and not ax64:
-            return 1
+        return 1
+    
+    for k in ["chip", "machineArch", "productArch"]:
+        r = archOrd(k, a) - archOrd(k, b)
+        if r != 0:
+            return r
+
     if "language" in a and "language" in b:
         aeng = a["language"].lower().startswith("en-")
         beng = b["language"].lower().startswith("en-")
@@ -234,7 +244,7 @@ def prioritizePackage(a, b):
             return 1
     return 0
 
-def getPackages(manifest):
+def getPackages(manifest, arch):
     packages = {}
     for p in manifest["packages"]:
         id = p["id"].lower()
@@ -242,7 +252,7 @@ def getPackages(manifest):
             packages[id] = []
         packages[id].append(p)
     for key in packages:
-        packages[key] = sorted(packages[key], key=functools.cmp_to_key(prioritizePackage))
+        packages[key] = sorted(packages[key], key=functools.cmp_to_key(functools.partial(prioritizePackage, arch)))
     return packages
 
 def listPackageType(packages, type):
@@ -640,7 +650,26 @@ if __name__ == "__main__":
 
     socket.setdefaulttimeout(15)
 
-    packages = getPackages(getManifest(args))
+    if args.host_arch is None:
+        args.host_arch = platform.machine().lower()
+        if platform.system() == "Darwin":
+            # There is no prebuilt arm64 Wine on macOS.
+            args.host_arch = "x64"
+        elif args.host_arch in ["x86", "i386", "i686"]:
+            args.host_arch = "x86"
+        elif args.host_arch in ["x64", "x86_64", "amd64"]:
+            args.host_arch = "x64"
+        elif args.host_arch in ["arm64", "aarch64"]:
+            args.host_arch = "arm64"
+        else:
+            args.host_arch = None
+
+    if args.host_arch is None:
+        print("WARNING: Unable to detect host architecture")
+    else:
+        print("Install packages for %s host architecture" % args.host_arch)
+
+    packages = getPackages(getManifest(args), args.host_arch)
 
     if args.print_version:
         sys.exit(0)
--- a/vsdownload.py
+++ b/vsdownload.py
@@ -60,7 +60,7 @@ def getArgsParser():
     return parser
 
 def setPackageSelectionMSVC16(args, packages, userversion, sdk, toolversion, defaultPackages):
-    if findPackage(packages, "Microsoft.VisualStudio.Component.VC." + toolversion + ".x86.x64", None, warn=False):
+    if findPackage(packages, "Microsoft.VisualStudio.Component.VC." + toolversion + ".x86.x64", warn=False):
         if sdk.startswith("10.0.") and int(sdk[5:]) >= 22000:
             sdkpkg = "Win11SDK_" + sdk
         else:
@@ -77,7 +77,7 @@ def setPackageSelectionMSVC16(args, packages, userversion, sdk, toolversion, def
         args.package.extend(defaultPackages)
 
 def setPackageSelectionMSVC15(args, packages, userversion, sdk, toolversion, defaultPackages):
-    if findPackage(packages, "Microsoft.VisualStudio.Component.VC.Tools." + toolversion, None, warn=False):
+    if findPackage(packages, "Microsoft.VisualStudio.Component.VC.Tools." + toolversion, warn=False):
         args.package.extend(["Win10SDK_" + sdk, "Microsoft.VisualStudio.Component.VC.Tools." + toolversion])
     else:
         # Options for toolchains for specific versions. The latest version in
@@ -268,7 +268,7 @@ def listPackageType(packages, type):
     for id in sorted(ids):
         print(id)
 
-def findPackage(packages, id, chip, warn=True):
+def findPackage(packages, id, constraints={}, warn=True):
     origid = id
     id = id.lower()
     candidates = None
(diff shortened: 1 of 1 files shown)


vsdownload.py
```diff
@@ -626,7 +639,25 @@ def moveVCSDK(unpack, dest):
 
     socket.setdefaulttimeout(15)
 
-    packages = getPackages(getManifest(args))
+    if args.host_arch is None:
+        args.host_arch = platform.machine().lower()
+        if platform.system() == 'Darwin':
+            args.host_arch = 'x64'
```
mstorsjo, line 646: The special treatment of Darwin here probably requires spelling out the reasons, so that we potentially can revisit it in the future.

I presume the reason is that native arm64 Wine on aarch64 doesn't really work on macOS yet, so we prefer x86_64 Wine in Rosetta over broken Wine on aarch64?

vsdownload.py
```diff
@@ -203,14 +205,25 @@ def getManifest(args):
...
+        c = c.lower()
+        if c == "neutral":
             return -1
-        elif bx64 and not ax64:
-            return 1
+        if c == arch:
+            return -2
+        return 1
+    
+    for k in ["chip", "machineArch", "productArch"]:
```
mstorsjo, line 222: Can you briefly explain what roles these fields, `chip`, `machineArch` and `productArch` have wrt the tools? Are they all about the host architecture for where the tools are going to run, or is some aspect of these also about the intended target architecture they're producing code for?

mstorsjo (commented):
I think this looks reasonable, but I'd like to understand the roles of these fields a little better (it's been ages since I looked closer at the package trees).

mstorsjo:
Does this change mean that we'll only end up downloading one of the `Hostx86`, `Hostx64` and `Hostarm64` directories for the compiler? That'd be a large size saving.

hacksysteam:
Hi @huangqinjin , We need to add/handle constraints to few more places, because it was passing `None`

```diff
diff --git a/vsdownload.py b/vsdownload.py
index 0e8e021..0f2852d 100755
--- a/vsdownload.py
+++ b/vsdownload.py
@@ -63,7 +63,7 @@ def getArgsParser():
     return parser
 
 def setPackageSelectionMSVC16(args, packages, userversion, sdk, toolversion, defaultPackages):
-    if findPackage(packages, "Microsoft.VisualStudio.Component.VC." + toolversion + ".x86.x64", None, warn=False):
+    if findPackage(packages, "Microsoft.VisualStudio.Component.VC." + toolversion + ".x86.x64", {}, warn=False):
         if sdk.startswith("10.0.") and int(sdk[5:]) >= 22000:
             sdkpkg = "Win11SDK_" + sdk
         else:
@@ -82,8 +82,9 @@ def setPackageSelectionMSVC16(args, packages, userversion, sdk, toolversion, def
         print("Didn't find exact version packages for " + userversion + ", assuming this is provided by the default/latest version")
         args.package.extend(defaultPackages)
 
+
 def setPackageSelectionMSVC15(args, packages, userversion, sdk, toolversion, defaultPackages):
-    if findPackage(packages, "Microsoft.VisualStudio.Component.VC.Tools." + toolversion, None, warn=False):
+    if findPackage(packages, "Microsoft.VisualStudio.Component.VC.Tools." + toolversion, {}, warn=False):
         args.package.extend(["Win10SDK_" + sdk, "Microsoft.VisualStudio.Component.VC.Tools." + toolversion])
     else:
         # Options for toolchains for specific versions. The latest version in
@@ -699,7 +700,7 @@ if __name__ == "__main__":
         sys.exit(0)
 
     if not args.accept_license:
-        response = input("Do you accept the license at " + findPackage(packages, "Microsoft.VisualStudio.Product.BuildTools", None)["localizedResources"][0]["license"] + " (yes/no)? ")
+        response = input("Do you accept the license at " + findPackage(packages, "Microsoft.VisualStudio.Product.BuildTools", {})["localizedResources"][0]["license"] + " (yes/no)? ")
         while response != "yes" and response != "no":
             response = input("Do you accept the license? Answer \"yes\" or \"no\": ")
         if response == "no":
```

vsdownload.py
huangqinjin: The reason is simple: there is no prebuilt arm64 Wine on macOS. The package maintainer states (https://github.com/Gcenx/macOS_Wine_builds/issues/74#issuecomment-1858738680):
> There no point build wine for aarch64.

mstorsjo: > The reason is simple: there is no prebuilt arm64 Wine on macOS.

Yep. (I experimented with building wine for arm64 macOS when this was in developer preview - I am familiar with the issues and challenges around it.)

> The package maintainer states ([Gcenx/macOS_Wine_builds#74 (comment)](https://github.com/Gcenx/macOS_Wine_builds/issues/74#issuecomment-1858738680)):
> 
> > There no point build wine for aarch64.

As the rest of the thread shows - that’s from the point of view if you want to run x86 windows binaries. For running aarch64 windows binaries, it would be entirely reasonable to do.

But the issue indeed is that it’s not available at the moment; mention this in a comment.

vsdownload.py
huangqinjin: In fact, I can't. I think these fields are all about the architecture of binaries in the packages, and different packages use different fields. But we can't know whether **the package is for host machine or target machine**. See https://github.com/mstorsjo/msvc-wine/issues/135#issuecomment-2371844447.

huangqinjin:
> Does this change mean that we'll only end up downloading one of the `Hostx86`, `Hostx64` and `Hostarm64` directories for the compiler? That'd be a large size saving.

No, this change just gives `"chip", "machineArch", "productArch"` a better default if they are not explicitly specified in `dependencies`.

huangqinjin:
> Hi @huangqinjin , We need to add/handle constraints to few more places, because it was passing `None`

Thanks. Amended.

vsdownload.py
huangqinjin: Done.

vsdownload.py
```diff
@@ -626,7 +639,26 @@ def moveVCSDK(unpack, dest):
 
     socket.setdefaulttimeout(15)
 
-    packages = getPackages(getManifest(args))
+    if args.host_arch is None:
+        args.host_arch = platform.machine().lower()
+        if platform.system() == 'Darwin':
+            # There is no prebuilt arm64 Wine on macOS.
+            args.host_arch = 'x64'
```
mstorsjo, line 647: Sorry to ask about this now, not having noted it before... We're using `""` for string literals in this script so far, almost consistently, so I'd appreciate if you'd match that style for the newly added code too.

hacksysteam: @mstorsjo i already fixed this in the PR i sent. This is what I was talking about normalizing the quotes

mstorsjo: I know. But different things belong in different commits. You should not mix in unrelated cleanups in a functional change.

Is this already has been merged to master, I would have pushed a commit to unify it, but as this is in a yet unmerged PR, I prefer fixing up within this PR before merging it.

hacksysteam: Got it. Once this is merged, I'll rebase and update the PR and include the changes you suggested

hacksysteam: Any plans to merge this sooner https://github.com/mstorsjo/msvc-wine/pull/144

This is very nice feature and has reduced the size of the container drastically.

vsdownload.py
huangqinjin: OK. We should treat them as host architecture. User can use `--host-arch` to download any architecture they want, no matter for host machine or target machine. So maybe `--machine-arch` is a better name? @mstorsjo any preference?

vsdownload.py
huangqinjin: > Sorry to ask about this now, not having noted it before... We're using `""` for string literals in this script so far, almost consistently, so I'd appreciate if you'd match that style for the newly added code too.

Done.

vsdownload.py
mstorsjo: I think `--host-arch` is a name that is easier to understand for most users - and in the vast majority of cases, users don't need to set the option at all. (And for the cases where you do want to set it, I guess it can be understood as "I want to download tools for host X", not necessarily "I'm running on host X right now".) So I think the current option name is fine - unless we at a later point find a reason why we should change it.

mstorsjo:
Thanks, let's merge this now!

vsdownload.py
huangqinjin: Fair enough.

PR #144, opened 2024-09-23
WIP: Add options for skipping downloading tools targeting ARM/ARM64, and for skipping ATL and DIA SDK

mstorsjo:
This goes on top of #143.

This works, but is quite primitive. If someone wants to pick this PR up and develop it further, that would be very much appreciated! (CC @huangqinjin)

I wanted to make it possible to skip downloading architectures, and/or extra libraries that you don't need. (Although ATL seems very small so it doesn't make much of a difference in any way whether we include it or not.)

I tried to make this general enough that you'd be able to say e.g. that you only want to target e.g. x86 and x64, and thus not adding the extra e.g. `*.ARM64` package selections. But if one only wants a compiler to target ARM but not x86/x64 at all, I think we'd need to manually select some more packages, at least for the latest version (when not selecting a version with `--msvc-version`).

As an extra, if one has skipped installing ARM/ARM64 files, we could also remove a lot of large libraries under WinSDK, but I didn't implement that yet.

### DIFF
--- a/vsdownload.py
+++ b/vsdownload.py
@@ -54,6 +54,9 @@ def getArgsParser():
     parser.add_argument("--keep-unpack", const=True, action="store_const", help="Keep the unpacked files that aren't otherwise selected as needed output")
     parser.add_argument("--msvc-version", metavar="version", help="Install a specific MSVC toolchain version")
     parser.add_argument("--sdk-version", metavar="version", help="Install a specific Windows SDK version")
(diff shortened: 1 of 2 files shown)


hacksysteam:
@mstorsjo much-needed feature. Most of the time, we don't need ARM and ARM64 architecture support and this causes the huge size of the container

PR #145, opened 2024-09-24
Improve output readability and download efficiency

hacksysteam:
* Merged https://github.com/mstorsjo/msvc-wine/pull/140
* Merged https://github.com/mstorsjo/msvc-wine/pull/144
* Added colors to output messages to highlight important information.
* Switched from multiprocessing to ThreadPoolExecutor for downloads.
* Added tree structure to `--print-deps`, `--print-reverse-deps` and `--print-selection`.

### DIFF
--- a/vsdownload.py
+++ b/vsdownload.py
@@ -54,6 +54,9 @@ def getArgsParser():
     parser.add_argument("--keep-unpack", const=True, action="store_const", help="Keep the unpacked files that aren't otherwise selected as needed output")
     parser.add_argument("--msvc-version", metavar="version", help="Install a specific MSVC toolchain version")
     parser.add_argument("--sdk-version", metavar="version", help="Install a specific Windows SDK version")
+    parser.add_argument("--architecture", metavar="arch", help="Target architecture to include (x86, x64, arm, arm64)", nargs="*")
+    parser.add_argument("--skip-atl", const=True, action="store_const", help="Skip installing the ATL headers")
+    parser.add_argument("--skip-diasdk", const=True, action="store_const", help="Skip installing the DIA SDK")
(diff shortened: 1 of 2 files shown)


hacksysteam:
To check the color output you can check: https://github.com/mstorsjo/msvc-wine/actions/runs/11022479747/job/30611873227?pr=145

you should also check the output from `--print-deps`, `--print-reverse-deps` and `--print-selection`. it looks nice. feedbacks are welcome

mstorsjo:
Sorry, but this changes way too much at once to be practically reviewable.

As you do essentially three different changes, I'd suggest splitting them up in three separate commits. If they're independent and don't clobber each other, they can go in separate PRs in parallel, otherwise they can go in separate commits within the same PR, or just do one of them in a PR, then post the next one once that is merged.

I also see that in addition to all other changes, this change intermixedly changes the whole script to use `'` for quotes rather than `"`. Don't mix such changes with functional changes. If there's a reason to change style that can be agreed upon, then it has to be done in a separate commit, separate from all functional changes.

hacksysteam:
Hi @mstorsjo thanks for the comments.

As you can see that this PR has multiple commits. Two merged PRs and regarding `'` and `"` I normalized the code to use `" ` as this is what being used in most of the places in the code.

Let me know how you want me to proceed and I'll update the PR accordingly.

mstorsjo:
> Hi @mstorsjo thanks for the comments.
> 
> As you can see that this PR has multiple commits. Two merged PRs and regarding `'` and `"` I normalized the code to use `" ` as this is what being used in most of the places in the code.

Yes, I know your PR contains multiple commits, for the other PRs. But your own commit on top consists of 3 different functional changes, plus the fact that you’re normalizing the quote style for strings. Don’t normalize anything, keep untouched code as it was. Split your 3 separate functional changes into 3 separate commits.

PR #148, opened 2024-10-05
vsdownload: Add an option to skip packages that do not match host arch

huangqinjin:
Depends on #140.

#### Some packages have their host arch in their ids, e.g.
```
"Microsoft.VisualCpp.Tools.HostARM64.TargetX64"
"Microsoft.VisualCpp.Tools.HostARM64.TargetX86"
"Microsoft.VisualCpp.Tools.HostX64.TargetX64"
"Microsoft.VisualCpp.Tools.HostX64.TargetX86"
```

I think it is pretty safe to skip packages if their ids do not match host arch.

[PR #140](https://github.com/mstorsjo/msvc-wine/actions/runs/11194999176/job/31122154650?pr=140#step:4:49)
```
Selected 544 packages, for a total download size of 2.9 GB, install size of 8.9 GB
```

[PR #148](https://github.com/mstorsjo/msvc-wine/actions/runs/11195105784/job/31122372211?pr=148#step:4:49)
```
Selected 475 packages, for a total download size of 2.7 GB, install size of 8.5 GB
```

### DIFF
--- a/vsdownload.py
+++ b/vsdownload.py
@@ -57,6 +57,7 @@ def getArgsParser():
     parser.add_argument("--sdk-version", metavar="version", help="Install a specific Windows SDK version")
     parser.add_argument("--with-wdk-installers", metavar="dir", help="Install Windows Driver Kit using the provided MSI installers")
     parser.add_argument("--host-arch", metavar="arch", choices=["x86", "x64", "arm64"], help="Specify the host architecture of packages to install")
+    parser.add_argument("--only-host", default=True, const=True, action="store_const", help="Only download packages that match host arch")
     return parser
 
 def setPackageSelectionMSVC16(args, packages, userversion, sdk, toolversion, defaultPackages):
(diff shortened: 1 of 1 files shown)


mstorsjo:
Thanks, this looks like a nice addition! However for some cases I guess it would be nice to be able to unpack all of them; would it be possible to add an option that lets the user get all of them?

huangqinjin:
A option is reasonable to me. But the main reason for me is that such filters based on host arch will going to be wrong in future, e.g. I want to filter packages based on their `chip`, `machineArch` and `productArch`.

Is `--only-host` a good name? What is the default value we want?

mstorsjo:
> A option is reasonable to me. But the main reason for me is that such filters based on host arch will going to be wrong in future, e.g. I want to filter packages based on their `chip`, `machineArch` and `productArch`.
> 
> Is `--only-host` a good name? What is the default value we want?

This looks like a reasonable name, and yes, we want the default to be to only unpack things for the host arch - as normal users won't need to think about this option. Thanks!

mstorsjo (approved):
LGTM

PR #149, opened 2024-10-06
vsdownload: Filter packages based on their arch information

huangqinjin:
Depends on #148.

#### Match host arch with `chip`, `machineArch` and `productArch`.
[PR #148](https://github.com/mstorsjo/msvc-wine/actions/runs/11198692222/job/31130172974?pr=148#step:4:49)
```
Selected 475 packages, for a total download size of 2.7 GB, install size of 8.5 GB
```

[PR #149 matching host arch exactly](https://github.com/mstorsjo/msvc-wine/actions/runs/11258538861/job/31305372957?pr=149)
```
Selected 449 packages, for a total download size of 2.5 GB, install size of 8.2 GB
```

**EDIT**: We go *matching host arch exactly*.

~~[PR #149](https://github.com/mstorsjo/msvc-wine/actions/runs/11202379694/job/31138414515?pr=149#step:4:49)~~

~~Selected 463 packages, for a total download size of 2.6 GB, install size of 8.3 GB~~

~~Current implementation apply the knownledge that~~
1. x64 host can run x86 binaries
1. arm64 host can run x86/x64 binaries

Otherwise some packages will be incomplete. E.g. package `Microsoft.VisualStudio.Debugger.Remote` on ARM64:
```shell
./vsdownload.py --accept-license --host-arch arm64 --print-deps-tree Microsoft.VisualStudio.Debugger.Remote

Microsoft.VisualStudio.Debugger.Remote
  Microsoft.VisualStudio.Debugger.Remote.Resources (chip.x64)
    Microsoft.VisualStudio.Debugger.Remote.Resources (chip.x86)
  Microsoft.VisualStudio.Debugger.Remote (chip.x86)
    Microsoft.VisualStudio.Debugger.Remote.Resources (chip.x86)
    Microsoft.VisualStudio.Debugger.Concord.Remote (chip.x86)
      Microsoft.VisualStudio.Debugger.Concord.Remote.Resources (chip.x86)
    Microsoft.VisualStudio.Debugger.Remote.DbgHelp.Win8 (chip.x86)
    Microsoft.VisualStudio.Debugger.Remote.ARM64
      Microsoft.VisualStudio.Debugger.Remote.Resources.ARM64
      Microsoft.VisualStudio.Debugger.Remote.ARM (machineArch.ARM64)
        Microsoft.VisualStudio.Debugger.Remote.Resources.ARM
        Microsoft.VisualStudio.Debugger.Concord.Remote.ARM (machineArch.ARM64)
          Microsoft.VisualStudio.Debugger.Concord.Remote.Resources.ARM
      Microsoft.VisualStudio.Debugger.Concord.Remote.ARM64 (machineArch.ARM64)
        Microsoft.VisualStudio.Debugger.Concord.Remote.Resources.ARM64
  Microsoft.VisualStudio.Debugger.Remote.DbgHelp.Win8 (chip.x64)
  Microsoft.VisualStudio.Debugger.Concord.Remote (chip.x64)
    Microsoft.VisualStudio.Debugger.Concord.Remote.Resources
```

But I think *matching host arch exactly* is also acceptable, and user need to specify `Microsoft.VisualStudio.Debugger.Remote.ARM64` instead of `Microsoft.VisualStudio.Debugger.Remote`.

### DIFF
--- a/vsdownload.py
+++ b/vsdownload.py
@@ -303,6 +303,13 @@ def matchPackageHostArch(p, host):
         if "host" + a in id:
             return a == host
 
+    for k in ["chip", "machineArch", "productArch"]:
+        a = p.get(k, "neutral").lower()
+        if a == "neutral":
+            continue
+        if a != host:
+            return False
+
     return True
 
 def printDepends(packages, target, constraints, indent, args):


vsdownload.py
```diff
@@ -303,6 +303,17 @@ def matchPackageHostArch(p, host):
         if "host" + a in id:
             return a == host
 
+    # Here we apply the knownledge that
+    # - x64 host can run x86 binaries
+    # - arm64 host can run x86/x64 binaries
```
mstorsjo, line 308: Hmm - I don't think we in general can assume that arm64 hosts can run x86 or x64 binaries? That's generally not the case, or what do you have in mind?

(In particular, even if you'd be running in an environment where you'd have something like Rosetta or qemu set up to emulate x86 binaries on arm64, you'd still need to have a suitable wine setup for the foreign architecture - which I wouldn't assume is the case in general.)

huangqinjin: > I don't think we in general can assume that arm64 hosts can run x86 or x64 binaries? That's generally not the case

It is the case for Windows. So it it possible that in the manifest, some arm64 packages depend on x64 packages and depend on x86 packages.

huangqinjin:
The example in the description is actually not correct. `Microsoft.VisualStudio.Debugger.Remote` on ARM64 have the properties
```
			"chip": "x64",
			"productArch": "neutral",
			"machineArch": "arm64",
```
So this package is not for pure ARM64. So anyway `Microsoft.VisualStudio.Debugger.Remote.ARM64` is needed for a complete `ARM64 Remote Debugger`.

So I am going the implement the filter *matching host arch exactly*.

huangqinjin:
[matching host arch exactly](https://github.com/mstorsjo/msvc-wine/actions/runs/11237902509/job/31241389016?pr=149#step:4:49)
```
```

vsdownload.py
mstorsjo: Right, I see what you mean.

vsdownload.py
```diff
@@ -303,6 +303,13 @@ def matchPackageHostArch(p, host):
         if "host" + a in id:
             return a == host
 
+    for k in ["chip", "machineArch", "productArch"]:
+        a = p.get(k, "neutral").lower()
+        if a == "neutral":
+            continue
+        if a != host:
+        	return False
```
mstorsjo, line 311: This line seems incorrectly indented

mstorsjo (commented):
The changes look good overall, except for the cosmetic issue

vsdownload.py
huangqinjin: No idea why a TAB is inserted. Amended.

mstorsjo (approved):
LGTM

Issue #114, opened 2024-02-15
CMake Find Packages


PatchByte:
How to use packages such as Vulkan.
Like if you would want to do:
```
find_package(Vulkan REQUIRED)
```
The problem is cmake would try to search after the linux package. I have not found any way to fix this.

mstorsjo:
When cross compiling with CMake, you need to set a bunch of cmake flags to tell it where/how to look for things, and what not to look for. One way of configuring those flags, which might work, is something like this:
```
-DCMAKE_FIND_ROOT_PATH=$CROSS_ROOT
-DCMAKE_FIND_ROOT_PATH_MODE_PROGRAM=NEVER
-DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=ONLY
-DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=ONLY
-DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ONLY
```

PR #150, opened 2024-10-11
Integrate with vcpkg

### LINKED ISSUE #141: CMake vcpkg refuses to work
Hi!

I'm getting this error when attempting to use vcpkg in CMake with the way described in the readme:
```
-- Running vcpkg install
error: in triplet x64-windows: Use of Visual Studio's Developer Prompt is unsupported on non-Windows hosts.
Define 'VCPKG_CMAKE_SYSTEM_NAME' or 'VCPKG_CHAINLOAD_TOOLCHAIN_FILE' in the triplet file.
-- Running vcpkg install - failed
```

I assume some more configuration is needed to make this work. What could it be?

huangqinjin:
Setting up CMake for cross compilation is tedious, e.g. #114 and #132. And using [vcpkg](https://github.com/microsoft/vcpkg) will greatly simplify the work.

Wrapper `findstr` is needed for vcpkg to install PDBs:
https://github.com/microsoft/vcpkg/blob/2024.09.30/scripts/cmake/vcpkg_copy_pdbs.cmake#L25

Closes #141.


test/test-vcpkg.sh
```diff
@@ -0,0 +1,120 @@
...
+    -DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake
+    -DVCPKG_TARGET_TRIPLET=$ARCH-windows
+)
+
+# Vcpkg uses pwsh and dumpbin to copy dependencies into the output directory for executables.
+if command -v pwsh &>/dev/null; then
+    export PATH=${BIN}:$PATH
+else
+    CMAKE_ARGS+=(
+        -DVCPKG_APPLOCAL_DEPS=OFF
```
mstorsjo, line 73: Can you explain this if/else a bit more? If we don't have pwsh, we pass `-DVCPKG_APPLOCAL_DEPS=OFF` so we skip the part that requires `pwsh`? And if we run the `pwsh` command, we also end up needing `dumpbin` and `findstr`, which is why we add `$BIN` to `$PATH` here?

huangqinjin: Yes, When building a user project with the vcpkg toolchain `vcpkg.cmake`, if `VCPKG_APPLOCAL_DEPS=ON`, which is the default, a post build command `applocal.ps1` is invoked
https://github.com/microsoft/vcpkg/blob/2024.09.30/scripts/buildsystems/vcpkg.cmake#L605-L619.

In the powershell script, `dumpbin`, `llvm-objdump` or `objdump` is used to get dll dependencies of the built binary. `findstr` is not used when building a user project, it is only used when `vcpkg` installing packages.
https://github.com/microsoft/vcpkg/blob/2024.09.30/scripts/buildsystems/msbuild/applocal.ps1#L96-L104

mstorsjo:
Thanks!

mstorsjo:
The CI test here seems to be failing now once this is merged into master, please have a look: https://github.com/mstorsjo/msvc-wine/actions/runs/11931942872

huangqinjin:
Vcpkg's PDB installation function is broken now, #154 disable related tests.

mstorsjo:
> Vcpkg's PDB installation function is broken now, #154 disable related tests.

Thanks for the quick followup!

PR #151, opened 2024-11-15
msbuild: Set UseEnv variable to true

TheComputerGuy96:
For some reason default Windows headers and libraries can't be found in an older project without those variables being set

### DIFF
--- a/wrappers/msbuild
+++ b/wrappers/msbuild
@@ -55,6 +55,8 @@ export SignMode=off
 export Inf2CatNoCatalog=true
 # API Validator crashes in WINE.
 export ApiValidator_Enable=False
+# Make sure the custom INCLUDE/LIB variables work.
+export UseEnv=true
 
 # Set MSBuild-specific Platform property via an environment variable.
 # 'Platform' property names do not match ARCH names exactly.


mstorsjo (approved):
LGTM

@ravenexp - Do you have any opinions on this?

ravenexp:
> @ravenexp - Do you have any opinions on this?

These variables look like something the MSVC property files for C++ projects would set internally.
See for example `MSBuild/Microsoft/VC/v170/Microsoft.Cpp.MSVC.Toolset.Win32.props`.
 
Rather than overriding these globally, I'd first check if setting `UseEnv` solves the problem for the project in question:

https://learn.microsoft.com/en-us/cpp/build/reference/msbuild-visual-cpp-overview?view=msvc-170#useenv-property

@TheComputerGuy96

Does
```sh
msbuild /p:UseEnv=true
```
work for your project?

TheComputerGuy96:
@ravenexp `UseEnv=true` also works (so I set that by default instead)

ravenexp:
> @ravenexp `UseEnv=true` also works (so I set that by default instead)

It's great that it works, but it should not be set by default because it effectively bypasses all configuration settings from the MSBuild project file.

Can you share your MSBuild project file or maybe just the configuration properties part without the source file names? I'll try to figure out what is missing from the msvc-wine MSBuild environment. I want to try to get to the root cause of this issue.

TheComputerGuy96:
@ravenexp Here's the .vcxproj file: https://github.com/CookiePLMonster/SilentPatch/blob/dev/DDraw/DDraw.vcxproj

ravenexp:
Ah, this must be it:

```xml
<PlatformToolset>v141_xp</PlatformToolset>
```

I've never tried targeting Windows 7, much less XP. I'll try to figure out what env vars are required to support this platform.

TheComputerGuy96:
I'm currently setting `/p:PlatformToolset=v143` option to make that project build (and the `SetEnv` option of course)

ravenexp:
> I'm currently setting `/p:PlatformToolset=v143` option to make that project build (and the `SetEnv` option of course)

I've opened https://github.com/mstorsjo/msvc-wine/pull/152 to address the root cause of this bug. With it applied, your project should now build with:

```sh
msbuild /p:Platform=Win32 /p:PlatformToolset=v143 /p:WindowsTargetPlatformVersion=10.0
```

`UseEnv` is not required.

TheComputerGuy96:
@ravenexp I can confirm that MR solves the issue too (with that extra `/p:WindowsTargetPlatformVersion=10.0` option; otherwise I get the same build error)

TheComputerGuy96:
Superseeded by #152

PR #152, opened 2024-11-16
wrappers(msbuild): Define the SDK search root paths

ravenexp:
Define yet more environment variables to make the platform SDK search methods work without the Windows Registry.


mstorsjo (approved):
LGTM, thanks!

PR #155, opened 2024-11-20
Add double-quotes to shell scripts (update)

TheComputerGuy96:
This is a rebased and minified version of #39 (which got stalled because of an extra script)


mstorsjo (approved):
Looks reasonable, thanks! I can merge this if the CI ends up succeeding.

PR #156, opened 2024-11-30
vsdownload: Fix unable to set `--only-host` to False

huangqinjin:
Previously `--only-host` didn't accept any argument. Now user can explicitly specify `yes` or `no`.

```shell
$ ./vsdownload.py -h
--only-host [yes|no]  Only download packages that match host arch (default: yes)
```

### DIFF
--- a/vsdownload.py
+++ b/vsdownload.py
@@ -31,6 +31,32 @@ import urllib.request
 import zipfile
 
 def getArgsParser():
+    class OptionalBoolean(argparse.Action):
+        def __init__(self,
+                    option_strings,
+                    dest,
+                    default=None,
+                    help=None):
+            
+            if default is not None:
+                default_string = "yes" if default else "no"
+                if help is None:
+                    help = "default: " + default_string
+                else:
+                    help += " (default: %s)" % default_string
+
+            super().__init__(
+                option_strings=option_strings,
+                dest=dest,
+                nargs='?',
+                default=default,
+                choices=["yes", "no"],
+                help=help,
+                metavar="yes|no")
+
+        def __call__(self, parser, namespace, values, option_string=None):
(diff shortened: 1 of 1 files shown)


vsdownload.py
```diff
@@ -31,6 +31,32 @@
...
+                option_strings=option_strings,
+                dest=dest,
+                nargs='?',
+                default=default,
+                choices=["yes", "no"],
+                help=help,
+                metavar="yes|no")
+
+        def __call__(self, parser, namespace, values, option_string=None):
+            setattr(namespace, self.dest, values != "no")
```
mstorsjo, line 58: So contrary to `super().__call__()`, this also contains the logic to map any other value than `no` to `yes`. (What about handling cases like `false`?)

mstorsjo (approved):
This looks good to me, although my Python is a bit too rusty to follow the details of how the custom `Action` works here.

vsdownload.py
huangqinjin: It only accepts values specified by `choices` and `None`, so it also maps `None` to `yes`. What is your preference about the values, I feel good about any following: 
1. `yes|no`
2. `true|false`
3. `yes|no|true|false`

mstorsjo: Ah, right, that helps a lot.

In that case - I don't have a strong opinion; either `yes|no` or `true|false`, but not both. Not sure what is more idiomatic in this environment though - I'd leave it up to you, and I guess your preference is for `yes|no` as that's what you chose so far.

huangqinjin: > I guess your preference is for `yes|no` as that's what you chose so far.

Yes, exactly.

mstorsjo: Ok, let's go with this then. Thanks!

PR #158, opened 2024-11-30
vsdownload: Add an option to specify target architectures

huangqinjin:


vsdownload.py
```diff
@@ -88,12 +97,27 @@ def setPackageSelectionMSVC15(args, packages, userversion, sdk, toolversion, def
         args.package.extend(defaultPackages)
 
 def setPackageSelection(args, packages):
+    if not args.architecture or len(args.architecture) == 0:
+        args.architecture = ["x86", "x64", "arm", "arm64"]
```
huangqinjin, line 101: IMO default architecture to host-arch only is a better option.

vsdownload.py
```diff
@@ -55,21 +55,30 @@ def getArgsParser():
...
     parser.add_argument("--with-wdk-installers", metavar="dir", help="Install Windows Driver Kit using the provided MSI installers")
     parser.add_argument("--host-arch", metavar="arch", choices=["x86", "x64", "arm64"], help="Specify the host architecture of packages to install")
     parser.add_argument("--only-host", default=True, const=True, action="store_const", help="Only download packages that match host arch")
     return parser
 
 def setPackageSelectionMSVC16(args, packages, userversion, sdk, toolversion, defaultPackages):
     if findPackage(packages, "Microsoft.VisualStudio.Component.VC." + toolversion + ".x86.x64", warn=False):
+        if "x86" in args.architecture or "x64" in args.architecture:
+            args.package.append("Microsoft.VisualStudio.Component.VC." + toolversion + ".x86.x64")
+            args.package.append("Microsoft.VC." + toolversion + ".ASAN.X86")
```
mstorsjo, line 68: It looks like this is an addition to what packages we installed before, but I guess it's a very nice addition. Would be good to mention it explicitly though, that this also changes the sets of packages that gets installed (or single out that change into a quick PR before or after this one).

vsdownload.py
```diff
@@ -640,6 +670,7 @@ def unpackWin10WDK(src, dest):
 
 def extractPackages(selected, cache, dest):
     makedirs(dest)
+    makedirs(os.path.join(dest, "MSBuild"))
```
mstorsjo, line 673: I presume this is a leftover from testing, the same as #159?

vsdownload.py
```diff
@@ -175,7 +206,6 @@ def setPackageSelection(args, packages):
                     args.package.append(key)
                 else:
                     args.ignore.append(key)
-        p = packages[key][0]
```
mstorsjo, line 178: This looks unrelated?

As general commentery, I recently tried to use the `--sdk-version` option, and I think we should error out, if the option was specified but it didn't match any packages - as it's not entirely obvious in which form the version number needs to be passed (e.g. `10.0.17763` without the trailing `.0`). I could try to do such a PR separately.

vsdownload.py
```diff
@@ -88,12 +97,34 @@ def setPackageSelectionMSVC15(args, packages, userversion, sdk, toolversion, def
...
+        defaultPackages.append("Microsoft.VisualStudio.Component.VC.Tools.ARM")
+        defaultPackages.append("Microsoft.VisualStudio.Component.VC.ATL.ARM")
+    if "arm64" in args.architecture:
+        defaultPackages.append("Microsoft.VisualStudio.Component.VC.Tools.ARM64")
+        defaultPackages.append("Microsoft.VisualStudio.Component.VC.ATL.ARM64")
+    
+    # TODO
+    defaultPackages.append("Win11SDK_10.0.22621")
+
+    defaultPackages.extend([
```
mstorsjo, line 123: I presume you're adding these packages manually here, to compensate for things that are left out by installing `Microsoft.VisualStudio.Component.VC.Tools.x86.x64` rather than `Microsoft.VisualStudio.Workload.VCTools`?

vsdownload.py
```diff
@@ -88,12 +97,34 @@ def setPackageSelectionMSVC15(args, packages, userversion, sdk, toolversion, def
...
+        defaultPackages.append("Microsoft.VisualStudio.Component.VC.ATL")
+    if "arm" in args.architecture:
+        defaultPackages.append("Microsoft.VisualStudio.Component.VC.Tools.ARM")
+        defaultPackages.append("Microsoft.VisualStudio.Component.VC.ATL.ARM")
+    if "arm64" in args.architecture:
+        defaultPackages.append("Microsoft.VisualStudio.Component.VC.Tools.ARM64")
+        defaultPackages.append("Microsoft.VisualStudio.Component.VC.ATL.ARM64")
+    
+    # TODO
+    defaultPackages.append("Win11SDK_10.0.22621")
```
mstorsjo, line 121: Is this also needed to compensate for missing dependencies when not using `Microsoft.VisualStudio.Workload.VCTools`?

If it turns out to be complicated to handle these bits, we could just unconditionally keep installing `Microsoft.VisualStudio.Workload.VCTools`, for now, and land the rest of the unproblematic parts of this PR, and then work on making more of the x86 tools optional in a separate later PR?

vsdownload.py
```diff
@@ -88,12 +97,34 @@ def setPackageSelectionMSVC15(args, packages, userversion, sdk, toolversion, def
         args.package.extend(defaultPackages)
 
 def setPackageSelection(args, packages):
+    if not args.architecture:
+        if args.host_arch is not None:
+            args.architecture = [ args.host_arch ]
```
mstorsjo, line 102: Hmm. I'm a little divided on this; changing this will affect existing users that run `vsdownload.py` expecting to get all cross compilers installed. I can see the case that many users only expect the host arch though.

mstorsjo (commented):
Thanks for working on this!

vsdownload.py
mstorsjo: Actually, when poking around more here - I see that it seems like this line is totally unused and broken leftover from whenever I first implemented this, so we probably just should remove it, independently from these other changes.

vsdownload.py
huangqinjin: Yes. Will rebase on top of #159 after it is accepted.

huangqinjin:
OK. I will retain `Microsoft.VisualStudio.Workload.VCTools` and install all architectures by default, To make this PR "No Functional Change" as possible.

mstorsjo:
> OK. I will retain `Microsoft.VisualStudio.Workload.VCTools` and install all architectures by default, To make this PR "No Functional Change" as possible.

Nice, thanks! That makes it easier to proceed. :-)

I also thought we could accept the choices e.g. `native` (or `host`) and `all`, as explicit shorthand choices. (Being able to do `--architectures all` is a bit easier than having to list all of them.)

mstorsjo (approved):
Looks good now, thanks!

PR #160, opened 2024-11-30
vsdownload: Allow dest dir to contain spaces on Windows

huangqinjin:


vsdownload.py
```diff
@@ -587,7 +587,8 @@ def unpackWin10SDK(src, payloads, dest):
             print("Extracting " + name)
             srcfile = os.path.join(src, name)
             if sys.platform == "win32":
-                cmd = ["msiexec", "/a", srcfile, "/qn", "TARGETDIR=" + os.path.abspath(dest)]
+                # The path to TARGETDIR need to be quoted in the case of spaces.
+                cmd = "msiexec /a \"%s\" /qn TARGETDIR=\"%s\"" % (srcfile, os.path.abspath(dest))
```
mstorsjo, line 591: Hmm, so is this a case where the python `subprocess.check_call()` taking an array `cmd` actually doesn't quote the individual arguments correctly when they contain spaces? (I know there are hairy corner cases when it comes to quoting of special characters, but for something as straightforward as regular spaces, I would have expected the usual array form to work.)

mstorsjo: Btw, for this change, it makes me think that it would be valuable to have CI testing using `vsdownload.py` on Windows too, to have test coverage of these codepaths. After downloading and unpacking, I wonder what trivial verification we could do on it though.

huangqinjin: > so is this a case where the python `subprocess.check_call()` taking an array `cmd` actually doesn't quote the individual arguments correctly when they contain spaces?

No. `msiexec` only accepts command line that in form of `msiexec TARGETDIR="a path"` and rejects `msiexec "TARGETDIR=a path"` (I don't know why!), though `argv` of these two form are same. The array form of `subprocess.check_call()` constructs the command line in the second form, which is quite standard, but is not suitable for `msiexec`.

mstorsjo: I see - that's weird, but I guess that's how it works. Thanks for the details!

mstorsjo:
I manually rebased this one now, to include the CI for running on Windows - will merge after seeing a successful run there.

PR #161, opened 2024-11-30
vsdownload: Unify the extract directories of msiextract and msiexec

huangqinjin:


vsdownload.py
```diff
@@ -656,22 +661,19 @@ def extractPackages(selected, cache, dest):
...
+    components = [
+        "VC",
+        "Windows Kits",
+        # The DIA SDK isn't necessary for normal use, but can be used when e.g.
+        # compiling LLVM.
+        "DIA SDK",
+        # MSBuild is the standard VC build tool.
+        "MSBuild",
+    ]
+    for dir in filter(None, components):
```
huangqinjin, line 675: Here ther filter is applied, so that in future we can do for instance 
```python
    "DIA SDK" if args.skip_diasdk else None,
```

mstorsjo (approved):
LGTM, I think this looks ok.

huangqinjin:
Rebased to trigger newly added Windows CI.

PR #162, opened 2024-12-03
vsdownload: Add option switches for each component

huangqinjin:
This PR introduces `--with-<component>` for each component,  and also `--with-default` as a shorthand to set all `--with-<component>` at once. `--with-*` has three possible values:
| value | description |
|-------|:-------------|
| yes   | the component is explicitly included |
| no    | the component is explicitly excluded |
|       | the component is not explicitly included but might be included through dependencies |

All `--with-*` have proper defaults, so that the selected packages are same with previous, for every combinations of command line options `msvc-version`, `sdk-version` and `package`. And we get finer-grained control over interested components by explicit `--with-<component>` options ,

| `msvc-version` | `sdk-version` | `package` | default | workload | msvc   | sdk    | dia    | msbuild |
|:--------------:|:-------------:|:---------:|:-------:|:--------:|:------:|:------:|:------:|:-------:|
|                |               |           | `True`  | I        | D -> I | D -> I | D -> I | D -> I  |
|                |               | `+`       | `None`  | N        | N      | N      | N      | N       |
| `a.b`          |               | `*`       | `None`  | N        | I      | I      | D      | N       |
| `a.b`          | `c.d.e`       | `*`       | `None`  | N        | I      | I      | D      | N       |
|                | `c.d.e`       |           | `True`  | I        | D -> I | I      | D -> I | D -> I  |
|                | `c.d.e`       | `+`       | `None`  | N        | N      | I      | N      | N       |

By `--with-default no`, it is achievable now to download a single Windows SDK package only.

| `msvc-version` | `sdk-version` | `package` | default | workload | msvc   | sdk    | dia    | msbuild |
|:--------------:|:-------------:|:---------:|:-------:|:--------:|:------:|:------:|:------:|:-------:|
|                | `c.d.e`       |           | no      | X        | X      | I      | X      | X       |

Some interested components do have `D -> I` change, it means, previously the component was included through dependencies, but now it is explicitly included. With these changes, it is possible to exclude workload to reduce download size, but we still get all interested components by default.

**Note:**
**I**: Explicitly included
**X**: Explicitly excluded
**D**: Included through dependencies
**N**: Not explicitly included but might be included through dependencies

### DIFF
--- a/vsdownload.py
+++ b/vsdownload.py
@@ -67,7 +67,7 @@ def getArgsParser():
     parser.add_argument("--cache", metavar="dir", help="Directory to use as a persistent cache for downloaded files")
     parser.add_argument("--dest", metavar="dir", help="Directory to install into")
     parser.add_argument("package", metavar="package", help="Package to install. If omitted, installs the default command line tools.", nargs="*")
-    parser.add_argument("--ignore", metavar="component", help="Package to skip", action="append")
+    parser.add_argument("--ignore", metavar="component", help="Package to skip", default=[], action="append")
     parser.add_argument("--accept-license", const=True, action="store_const", help="Don't prompt for accepting the license")
     parser.add_argument("--print-version", const=True, action="store_const", help="Stop after fetching the manifest")
     parser.add_argument("--list-workloads", const=True, action="store_const", help="List high level workloads")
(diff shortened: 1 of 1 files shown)


vsdownload.py
```diff
@@ -56,6 +82,11 @@ def getArgsParser():
     parser.add_argument("--msvc-version", metavar="version", help="Install a specific MSVC toolchain version")
     parser.add_argument("--sdk-version", metavar="version", help="Install a specific Windows SDK version")
     parser.add_argument("--architecture", metavar="arch", help="Target architecture to include (x86, x64, arm, arm64), defaults to host", nargs="+")
+    parser.add_argument("--with-msvc", action=OptionalBoolean, help="Include MSVC build tools (default)")
+    parser.add_argument("--with-sdk", action=OptionalBoolean, help="Include Windows SDK (default)")
+    parser.add_argument("--with-atl", action=OptionalBoolean, help="Include ATL")
```
mstorsjo, line 87: I think that at least ATL is useful enough that we should keep installing it by default.

mstorsjo (commented):
Thanks for working on this! This change feels quite complicated, but I think I managed to understand most of what of the changes do here.

I think the main potentially controversial bit here is changing the defaults... Changing the default for the DIA SDK probably is fine, possibly for MSBuild too. But I think it may be worthwhile to keep ATL installed by default.

vsdownload.py
```diff
@@ -111,117 +126,176 @@ def setPackageSelectionMSVC16(args, packages, userversion, sdk, toolversion, def
...
-    defaultPackages = ["Microsoft.VisualStudio.Workload.VCTools"]
+    defaultPackages, args.package = args.package, []
+    defaultIgnores, args.ignore = args.ignore, []
+
+    appendPackageSelection(args, args.with_workload, "Microsoft.VisualStudio.Workload.VCTools")
+
     if "x86" in args.architecture or "x64" in args.architecture:
-        defaultPackages.append("Microsoft.VisualStudio.Component.VC.ATL")
+        appendPackageSelection(args, args.with_msvc, "Microsoft.VisualStudio.Component.VC.Tools.x86.x64")
+        appendPackageSelection(args, args.with_asan, "Microsoft.VisualCpp.ASAN.X86")
```
mstorsjo, line 182: If I understand the patch correctly, this ASAN package, and the x86.64 tools above, are packages which previously got pulled in automatically through `Workload.VCTools` above. Now in a total default run, `vsdownload.py --dest /opt/msvc --accept-license`, we'd get them explicitly pulled in both implicitly via the workload and explicitly - is that correct?

In that case, do I understand correctly, that we also could make `--with-workload` default to false, to reduce the default install size, as we now do explicitly select the individual packages that we do need? (I guess we could do that as a separate commit then.)

mstorsjo (commented):
Thanks for the PR, this looks quite good overall I think.

I wonder if it would be good to add some tests for this package selection logic. This should be possible to do quite easily and quickly in the same way as the example I showed in https://github.com/mstorsjo/msvc-wine/pull/193#issuecomment-2955560830; something like `./vsdownload.py --accept-license --print-selection --msvc-version 17.11 | grep Win..SDK`. With that, it should be possible to show and test concrete cases for various combinations of options, that we do end up installing roughly the expected packages. (For historical `--msvc-version` we should be able to check that we do get e.g. a close enough match for the packages, and e.g. no extra SDK installed. For potential future versions we'd probably want to check more loosely, that we do get e.g. correct looking compiler packages, and check e.g. that if we request omitting something, that we don't select those packages.)

Side note; AFAIK the next major MSVS version, 18, should be coming in preview sometime in the end of this year. Hopefully it doesn't deviate too much from the package naming style of 16/17 :-)

vsdownload.py
huangqinjin: Yes, you are correct. Making `--with-workload` default to false is the goal, but it need more tests to see if any more packages are needed.

vsdownload.py
```diff
@@ -111,117 +126,176 @@ def setPackageSelectionMSVC16(args, packages, userversion, sdk, toolversion, def
...
-
-    if args.sdk_version != None:
-        found = False
+    if args.with_sdk and args.sdk_version is None:
+        # get recommended SDKs from the manifest
+        package = args.package
+        include_optional = args.include_optional
+        skip_recommended = args.skip_recommended
+        
+        args.package = ["Microsoft.VisualStudio.Workload.VCTools"]
```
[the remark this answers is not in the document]
huangqinjin: Already on master.

mstorsjo:
I think this is fine now; I've been mostly ok with it before, and this seems to just be a rebase of it, plus some of the minor comments addressed. Adding tests for it, or otherwise finetuning things can probably be left for later; it's probably best to just merge this now - as you no longer consider this PR a draft.

mstorsjo:
Oh, I see that https://github.com/mstorsjo/msvc-wine/pull/196 adds functionality which may be useful for testing this - I had entirely missed that PR...

PR #165, opened 2024-12-30
github: Test running vsdownload.py on Windows

mstorsjo:
CC @huangqinjin

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -98,3 +98,15 @@ jobs:
           docker build -f Dockerfile.clang .
       # Intentionally not pushing the built docker images;
       # the installed files aren't redistributable!
+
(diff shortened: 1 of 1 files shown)


huangqinjin:
Thanks. this is also what I want to do. FYI, I have a WIP feature to install and patch VsDevCmd/vcvars, so that we can use them to setup MSVC environment, just like Developer Command Prompt provided by a installed VS. And I will enrich the tests on Windows.

mstorsjo:
> Thanks. this is also what I want to do. FYI, I have a WIP feature to install and patch VsDevCmd/vcvars, so that we can use them to setup MSVC environment, just like Developer Command Prompt provided by a installed VS. And I will enrich the tests on Windows.

Nice, that sounds useful! I'll go ahead and merge this then.

PR #166, opened 2024-12-30
vsdownload: Error out if the requested SDK version isn't found

mstorsjo:
Also list the potential alternatives.

Also remove a leftover unused line of code, that seems to have been included by accident in the commit that added this feature, in 23d7c7b817ea21deba6d15c76ad655ee31d0e2e6.

CC @huangqinjin

### DIFF
--- a/vsdownload.py
+++ b/vsdownload.py
@@ -22,6 +22,7 @@ import os
 import multiprocessing.pool
 import json
 import platform
+import re
 import shutil
 import socket
 import subprocess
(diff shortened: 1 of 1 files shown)


vsdownload.py
```diff
@@ -167,15 +168,26 @@ def setPackageSelection(args, packages):
...
+                    found = True
                     args.package.append(key)
                 else:
                     args.ignore.append(key)
-        p = packages[key][0]
+        if not found:
+            print("WinSDK version " + args.sdk_version + " not found")
+            print("Available versions:")
+            for v in versions:
+                print("    " + v)
```
huangqinjin, line 189: It could be better to sort `versions` explicitly print it out.

mstorsjo: Ok, done, thanks!

(It could perhaps also be nice to have a way to get these listed without having to mistype a version number first, but I'm not sure if that's worth the extra effort, doing e.g. `--sdk-version list` works just fine in this way too, even if the tool tells you it was an error.)

mstorsjo:
Whoops, I removed the wrong branch here accidentally...

PR #169, opened 2025-01-01
vsdownload: Install ASAN along with MSVC

huangqinjin:
ASAN is great. It must be included!

For versionless toolchains, package `Microsoft.VisualCpp.ASAN.X86` is already a dependency of `Microsoft.VisualStudio.Workload.VCTools`.


mstorsjo:
Nice, thanks! Can you amend the commit message to include the comment saying that this already is included in the versionless toolchains?

huangqinjin:
> Nice, thanks! Can you amend the commit message to include the comment saying that this already is included in the versionless toolchains?

Done.

PR #171, opened 2025-01-01
github: Add test coverage of --major, --msvc-version and --sdk-version

mstorsjo:
CC @huangqinjin

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -113,3 +113,50 @@ jobs:
         run: |
           ls "c:/msvc/Windows Kits/10/Include/*/um/winnt.h"
           ls "c:/msvc/VC/Tools/MSVC/*/include/intrin.h"
+
+  test-msvc-wine-versions:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install prerequisites
+        run: |
+          sudo dpkg --add-architecture i386 && sudo apt-get update
+          sudo apt-get install wine64 wine32 python3 msitools ca-certificates winbind
+          WINE=$(command -v wine64 || command -v wine || false)
+          $WINE wineboot
+      - uses: actions/checkout@v4
+      - name: Download MSVC 2017 latest
+        run: |
+          ./vsdownload.py --accept-license --dest $(pwd)/msvc --major 15
+          ./install.sh $(pwd)/msvc
+          ./msvc/bin/x64/cl test/hello.c -Fehello.exe
+          rm -rf msvc
+      - name: Download MSVC 2019 latest
+        run: |
+          ./vsdownload.py --accept-license --dest $(pwd)/msvc --major 16
+          ./install.sh $(pwd)/msvc
+          ./msvc/bin/x64/cl test/hello.c -Fehello.exe
+          rm -rf msvc
+      - name: Download MSVC 2017 specific version
+        run: |
+          ./vsdownload.py --accept-license --dest $(pwd)/msvc --major 15 --msvc-version 15.5
+          ./install.sh $(pwd)/msvc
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -110,3 +110,49 @@ jobs:
...
+          rm -rf msvc
+      - name: Download MSVC 2022 specific version
+        run: |
+          ./vsdownload.py --accept-license --dest $(pwd)/msvc --major 17 --msvc-version 17.0
+          ./install.sh $(pwd)/msvc
+          ./msvc/bin/x64/cl test/hello.c -Fehello.exe
+          rm -rf msvc
+      - name: Download MSVC 2022 with non-default SDK
+        run: |
+          ./vsdownload.py --accept-license --dest $(pwd)/msvc --major 17 --sdk-version 10.0.26100
```
huangqinjin, line 155: This version is newer than current default, so it is possible that it becomes the new default in future. A older version should be used.

huangqinjin (commented):
Any reason not to run full tests?

mstorsjo:
> Any reason not to run full tests?

One and a half reason; primarily to conserve running time. Right now, this tests 6 installations, and while running a minimal number of tests, it still takes as long as the regular linux-test-everything case. If we'd test everything, it would take up to 6 times as long. Secondarily, it's possible that these older versions don't work 100% in every corner case (things like how mspdbsrv.exe behaves etc, or wrt cmake/meson integration). It's of course possible that such an issue also is something that we might want to look into. While doing a minimal set of smoke tests we'd still make sure that the toolchain generally is usable, while saving a lot of runtime, and potentially avoiding corner cases that actually don't work fully.

.github/workflows/build.yml
mstorsjo: That's true, but this scenario is actually the primary reason why I added `--sdk-version` in the first place; to get the latest SDK when it's not pulled in by default. In `install.sh` we have a little bit of logic around which MSVC and WinSDK version to pick if there happens to be multiple version number directories - currently the logic is to pick the oldest MSVC version and the newest WinSDK version.

True - if this version becomes the default, this test somewhat loses the effectiveness, but it still covers manually picking a WinSDK version rather than the implicitly picked default.

I guess we could add a 7th case where we test intentionally installing an older SDK version; I considered doing that as part of the tests with `--msvc-version` too, but as `--msvc-version` also implicitly picks an SDK version via a different mechanism, I thought I'd keep that to be tested separately as well.

(As we don't explicitly check that we do get the actual MSVC/SDK version that we asked for here, the main value from the test comes from seeing that `vsdownload.py` doesn't fail and we manage to get a toolchain that can compile things overall, in the end.)

mstorsjo:
Also a separate observation here; when installing older versions like this, we sometimes get more than one versioned directory in `VC/Tools/MSVC` and/or in the WinSDK, and we have different logic for picking the right/intended one. This mostly works, at least for these cases I tried to test here.

But I also tested with e.g. `--msvc-version 15.8`, and that currently lands us with three different directories, `14.15.26706`, `14.15.26726` and `14.16.27023`. `14.15.26706` only contains a couple of (arm64) include files, while `14.15.26726` contains the bulk of the install (the intended one). `14.16.27023` also seems mostly complete, which seems to be the latest version within that major version.

So currently, `install.sh` doesn't figure out this particular case correctly - it's kinda messy. It would be nicer if we could avoid getting the extra, latest version of the compiler pulled in. (Passing more metadata from the `vsdownload.py` stage, about the intent of the install, also might help, but just avoiding installing too much also would fix most of the issue, I think.)

huangqinjin:
Yes. I also encountered some annoying corner cases when not installing the latest toolchains. So I think these smoke tests are enough for test coverage.

PR #173, opened 2025-01-08
Install VsDevCmd

huangqinjin:
1. Install Common7/Tools
This directory contains batch scripts to setup Developer Command Prompt.
Environment variable `VS170COMNTOOLS` points to this directory, and some
tools use it to locate VS installation path and MSVC toolchains.

2. Patch VsDevCmd
To make it lookup Windows SDK in the installation root.
Also workaround some Wine bugs.
Wine 10 (not yet released) is required to run `test-vcvars.sh`.
But `VsDevCmd.bat` may be useable on Wine 9 with some minor bugs.


vsdownload.py
```diff
@@ -723,6 +724,27 @@ def extractPackages(selected, cache, dest):
         else:
             print("Skipping unpacking of " + p["id"] + " of type " + type)
 
+def patchPackages(dest):
+    patches = os.path.join(os.path.dirname(os.path.abspath(__file__)), "patches")
+    for p in glob.glob("**", root_dir=patches, recursive=True):
```
huangqinjin, line 729: `root_dir` only available since python 3.10+.

vsdownload.py
```diff
@@ -839,6 +865,8 @@ def moveVCSDK(unpack, dest):
             moveVCSDK(unpack, dest)
             if not args.keep_unpack:
                 shutil.rmtree(unpack)
+            if not args.skip_patch:
+                patchPackages(dest)
```
huangqinjin, line 869: Only apply to latest VS (major 17 currently).

test/test-vcvars.sh
```diff
@@ -0,0 +1,127 @@
...
+EXEC "" TestUnixPath VCToolsInstallDir  $MSVCDIR
+EXEC "" TestUnixPath WindowsSdkDir      $SDKBASE
+EXEC "" TestVariable WindowsSDKVersion  $(. "${BIN}msvcenv.sh" && echo $SDKVER)
+EXEC "" TestUnixPath UniversalCRTSdkDir $SDKBASE
+EXEC "" TestUnixPath UCRTVersion        $(. "${BIN}msvcenv.sh" && echo $SDKVER)
+# Below tests require where.exe (available in Wine 9.3+).
+# EXEC "" TestUnixPath cl_exe             $(. "${BIN}msvcenv.sh" && echo $BINDIR)/cl.exe
+# EXEC "" TestUnixPath rc_exe             $(. "${BIN}msvcenv.sh" && echo $SDKBINDIR)/rc.exe
+# EXEC "" TestUnixPath MSBuild_exe        $(. "${BIN}msvcenv.sh" && echo $MSBUILDBINDIR)/MSBuild.exe
+
```
huangqinjin, line 126: Should conditionally enable these tests based on wine version, as Wine 10 is available on macOS.

vsdownload.py
```diff
@@ -723,6 +724,29 @@ def extractPackages(selected, cache, dest):
...
+    if not os.path.isdir(patches):
+        return
+    for patch in glob.iglob(os.path.join(patches, "**"), recursive=True):
+        if os.path.isdir(patch):
+            continue
+        p = os.path.relpath(patch, patches)
+        f, op = os.path.splitext(p)
+        if op == ".patch":
+            if os.access(os.path.join(dest, f), os.F_OK):
+                if 0 != subprocess.call(["git", "--work-tree=.", "apply", "--quiet", "--reverse", "--check", patch], cwd=dest):
```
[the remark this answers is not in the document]
huangqinjin: `--reverse` is used to check if the patch has already been applied. I added a comment above the line.

Changed to `if call() != 0:`,

huangqinjin:
I have extended README.md to describe the intended usages.

The classic way to use MSVC from command line is to launch one of Developer Command Prompt installed by VS. They just call one of the `vcvars` batch files (thin wrappers of `Common7\Tools\VsDevCmd.bat`) in `VC\Auxiliary\Build`.

Newer VS prefers to call `VsDevCmd.bat` directly though. Below is a Windows Terminal profile automatically added by VS.
ms-terminal-vsdevcmd-profile

huangqinjin:
These batch files are needed. For example vcpkg detects the existence of these files https://github.com/microsoft/vcpkg-tool/blob/2025-06-02/src/vcpkg/visualstudio.cpp#L85-L357, to check the supported target architectures.

Initially I thought we could rewrite these files in a simpler form like `msvcenv.sh`, as what they do is just to set some environment variables like `PATH`, `INCLUDE`, `LIB`. But I finally chose to patch them instead of rewrite them, as the batch files support many options, e.g. select msvc versions and sdk versions, and I would like to make `msvc-wine` a portable replacement of preinstalled VS. Also the actual patchset is small - less than 20 LOC.

I agree that including patch files is not a good option, they are not easy to read and update. But I think, as you said, it is the neatest way of solving the issue for now.

PR #174, opened 2025-01-10
github: Update `wine-mono` to v8.1.0 on Ubuntu 24.04

### LINKED ISSUE #172: wdk-download.sh fails on Ubuntu 24.04
@ravenexp 

The wine version available on Ubuntu 24.04 seems to fail when running `wdk-download.sh` - see e.g. https://github.com/mstorsjo/msvc-wine/actions/runs/12649755043/job/35246831703. To work around this, I've pinned the main Linux CI build job to run on `ubuntu-22.04` rather than `ubuntu-latest`, see 3b4e6b30db495d548b3d6745ac75d1073488fb7d. (Also see https://github.com/mstorsjo/msvc-wine/commit/7f1dcc2b80cf38d702e761eb8172a8245ac007de where I skip the WDK install, and the build otherwise works on 24.04.)

ravenexp:
Switch the Linux CI workflow back to `ubuntu-latest`.

Update `wine-mono` to version 8.1.0 as instructed in
https://gitlab.winehq.org/wine/wine/-/wikis/Wine-Mono#versions

Resolves #172

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,10 @@ env:
 
 jobs:
   test-msvc-wine-linux:
-    runs-on: ubuntu-latest
+    # Using a pinned version of Ubuntu here; this determines the version of
+    # Wine that gets installed - the version of wine-mono that is installed
+    # below depends on the host version of Wine.
+    runs-on: ubuntu-24.04
     steps:
       - name: Install prerequisites
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,8 +9,7 @@ env:
 
 jobs:
   test-msvc-wine-linux:
-    # wdk-download.sh doesn't work on ubuntu 24.04 (ubuntu-latest)
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - name: Install prerequisites
         run: |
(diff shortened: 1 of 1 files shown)


ravenexp:
Sorry for sitting on this for 3 days, it turned out to be a 10 minutes "Ah!" fix. I ended up over-allocating time for investigating the issue.

> As we rely on this to match the version from the host linux distribution, perhaps we should pin this job to ubuntu-24.04 so that we don't get another similar surprise in 2 years, but can bump things in a more controlled fashion then?

That makes sense. I only use `msvc-wine` for building custom Debian-based Docker containers, so the WINE distribution version is naturally pinned to the base container.

Unfortunately I don't know how to infer `wine-mono` version from `wine` package version in a script, so pinning seems to be the only choice.

mstorsjo:
> Sorry for sitting on this for 3 days, it turned out to be a 10 minutes "Ah!" fix. I ended up over-allocating time for investigating the issue.

No problem - there's really no rush here :-)

> > As we rely on this to match the version from the host linux distribution, perhaps we should pin this job to ubuntu-24.04 so that we don't get another similar surprise in 2 years, but can bump things in a more controlled fashion then?
> 
> That makes sense. I only use `msvc-wine` for building custom Debian-based Docker containers, so the WINE distribution version is naturally pinned to the base container.
> 
> Unfortunately I don't know how to infer `wine-mono` version from `wine` package version in a script, so pinning seems to be the only choice.

Yeah that's totally fine; especially as long as we have a link to the wine mono wiki nearby where it is pinned, it should be quite manageable.

mstorsjo:
I added a commit to this PR, for explicitly pinning the base ubuntu version here, with some more comments.

PR #177, opened 2025-02-25
vsdownload: Add --migrate to help with download cache refreshes

imaami:
I added a quick feature to `vsdownload.py` to help with download cache refreshes (e.g. on GitHub).

This is meant to help when you have a previous cache dir that might have accumulated old package files in it, and you need to create a new debloated cache. Use `--cache=newdir --migrate=olddir`  and the old location is used as a source for any needed files that may already exist. Existing files in the old cache are _moved_ to the new location, and any missing files are downloaded to the new location as usual.

The end result is that you can `rm -fr` the old cache dir and just move the new one in its place, no unneeded files left.

### DIFF
--- a/vsdownload.py
+++ b/vsdownload.py
@@ -64,6 +64,7 @@ def getArgsParser():
     parser.add_argument("--major", default=17, metavar="version", help="The major version to download (defaults to 17)")
     parser.add_argument("--preview", dest="type", default="release", const="pre", action="store_const", help="Download the preview version instead of the release version")
     parser.add_argument("--cache", metavar="dir", help="Directory to use as a persistent cache for downloaded files")
+    parser.add_argument("--migrate", metavar="dir", help="Old cache directory from which to migrate selected packages when downloading into a new cache")
     parser.add_argument("--dest", metavar="dir", help="Directory to install into")
     parser.add_argument("package", metavar="package", help="Package to install. If omitted, installs the default command line tools.", nargs="*")
     parser.add_argument("--ignore", metavar="component", help="Package to skip", action="append")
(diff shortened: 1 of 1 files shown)


imaami:
There's one thing I think I should fix: empty directories are left undeleted in the old cache.

imaami:
In my own tests this seems to now work as it should. This PR combined with #180 allow integration with https://github.com/marketplace/actions/cache. In my current workflow I store the downloaded payloads with a cache key based on their hashes, and when cache misses happen this PR (`--migrate`) makes it possible to recreate the new cache without redownloading any payloads that the old cache already has.

mstorsjo (commented):
I had a read through here, and it looks ok, I think.

It does touch a bit more code than I'd ideally see happening, and it adds a bit more code than I'd like for what I feel is an unusual configuration, but perhaps it isn't all that much in the overall picture.

Does @huangqinjin have any opinions on it?

I'm not sure if I saw any discussion about the motivation for it. In practice, I haven't really used the cache option at all since adding it. Is the motivation for using a cache when using msvc-wine in github actions to be a good netizen and not do unnecessary downloads when possible, or to speed up the installs? Do you have concrete numbers on how long it takes to install msvc-wine with and without cache, for reference?

And I guess the existing cache option would work as such, but it would cause the cache to continuously grow if not pruned (and we don't have any good logic for knowing what to prune - hence this PR, I guess). When using e.g. https://github.com/hendrikmuhs/ccache-action, such a cache also continuously grows, requiring one to set a max size for it. But there, ccache uses the max size to prune out things that are the least relevant...

PR #180, opened 2025-02-28
vsdownload: Add --print-hashes for download cache fingerprinting

imaami:
This is a simple feature, but very useful for caching the downloads. It simply prints all the SHA256 sums for the payloads in the selection.

This is a good way to reliably generate a key for a caching service like e.g. the GitHub cache action. All you need is the manifest and you can do the cache request with no ambiguity.


imaami:
Note to self: I think this still needs 3 changes:

1. Instead of printing out one line at a time on the go, accumulate the output and print it all at once when done. (I don't think there's any use in having an unbuffered "streaming" mode for this command.)
2. Pre-sort the list based on the paths, otherwise the output is not sufficiently reliable input for cache keying. Currently the ordering is dictated by implementation details of the tree search. Also, the sorting must be done as if `LC_ALL=C` is always set to account for locales being weird.
3. Nitpick: instead of 2 spaces, the hash and path should be separated by `" *"` to explicitly define the inpit files as binary data.

imaami:
Those 3 missing things I mentioned are now fixed. The output is sorted on the UTF-8-encoded paths. A given set of payloads will always produce the exact same output with `--print-hashes` (not accounting for possible CRLF).

imaami:
@mstorsjo I just rebased this and PR #177, the test-msvc-wine-macos failure seems unrelated to either PR. I checked and there's no difference that seems relevant. Nothing at all has changed in my edits since before the rebase, and all tests passed then. The only difference is the current head commit in master, 641cb48facd7fa273bc36cf718470097c433abd8.

Looking at the build logs there's one thing that sticks out. The runners that handled my PR branches are both an earlier runner image version. Maybe the older image needs mono installed and the newer doesn't?

Screenshots from a diff between the raw logs; the green is my PR and the red is master.

2025-03-02-160333_2596x602_scrot
2025-03-02-160236_1719x394_scrot

mstorsjo:
> @mstorsjo I just rebased this and PR #177, the test-msvc-wine-macos failure seems unrelated to either PR. I checked and there's no difference that seems relevant. Nothing at all has changed in my edits since before the rebase, and all tests passed then. The only difference is the current head commit in master, [641cb48](https://github.com/mstorsjo/msvc-wine/commit/641cb48facd7fa273bc36cf718470097c433abd8).
> 
> Looking at the build logs there's one thing that sticks out. The runners that handled my PR branches are both an earlier runner image version. The older image has mono 8.1.0 while the newer has 9.4.0.
> 
> Screenshots from a diff between the raw logs; the green is my PR and the red is master.
> 
> 2025-03-02-160333_2596x602_scrot 2025-03-02-160236_1719x394_scrot

Hmm, this is tricky if we still can end up getting either the older or newer versions - CC @huangqinjin.

(Overall, sorry I haven't been able to tend to your PRs yet; I've been away for a bit and am catching up on things. I'll try to get to a proper review of your submissions soon.)

imaami:
> > @mstorsjo I just rebased this and PR #177, the test-msvc-wine-macos failure seems unrelated to either PR. I checked and there's no difference that seems relevant. Nothing at all has changed in my edits since before the rebase, and all tests passed then. The only difference is the current head commit in master, [641cb48](https://github.com/mstorsjo/msvc-wine/commit/641cb48facd7fa273bc36cf718470097c433abd8).
> > Looking at the build logs there's one thing that sticks out. The runners that handled my PR branches are both an earlier runner image version. The older image has mono 8.1.0 while the newer has 9.4.0.
> > Screenshots from a diff between the raw logs; the green is my PR and the red is master.
> > 2025-03-02-160333_2596x602_scrot 2025-03-02-160236_1719x394_scrot
> 
> Hmm, this is tricky if we still can end up getting either the older or newer versions - CC @huangqinjin.
> 
> (Overall, sorry I haven't been able to tend to your PRs yet; I've been away for a bit and am catching up on things. I'll try to get to a proper review of your submissions soon.)

Don't worry about the PR merging timetables, none of this is urgent.

I found the reason, they actually reverted a recent macOS image update. This seems like a platform-wide thing, not arbitrary image changes during normal operation. https://github.com/actions/runner-images/pull/11700

imaami:
There's a solution: check the installed version of wine-mono.

```sh
    if ! somethingsomething; then
          curl -s -L -O https://github.com/madewokherd/wine-mono/releases/download/wine-mono-9.0.0/wine-mono-9.0.0-x86.msi
          $WINE msiexec /i wine-mono-9.0.0-x86.msi
    fi
```

I just don't know exactly how to do that.

huangqinjin:
> I found the reason, they actually reverted a recent macOS image update. This seems like a platform-wide thing, not arbitrary image changes during normal operation. [actions/runner-images#11700](https://github.com/actions/runner-images/pull/11700)

#183.

PR #182, opened 2025-03-01
Revert "github: Install Wine Mono 9.0 for macOS"

huangqinjin:
This reverts commit a40dd807ceaa189783766c447ea2c67b0d68f588.

Wine 10.0 bundled with Wine Mono 9.4.0, is now available on github macOS runners.


imaami:
I was wondering why I didn't need an explicit mono install command in my workflow, this explains it.

PR #190, opened 2025-05-16
Windows SDK 10.0.26100.0 no longer targets Arm32, skip related tests

huangqinjin:


huangqinjin:
@ravenexp I have no knowledge to solve https://github.com/mstorsjo/msvc-wine/actions/runs/15072745771/job/42372920934#step:5:816
> test.inx : error 1199: The syntax 'DIRID 13 (CopyFiles)' was introduced in OS version 10.0.16299, but DDInstall sections utilizing the syntax will install on earlier OS versions. Those DDInstall sections should be restricted to only install on 10.0.16299 or higher using a TargetOSVersion decoration.

Could you advise?

ravenexp:
> @ravenexp I have no knowledge to solve https://github.com/mstorsjo/msvc-wine/actions/runs/15072745771/job/42372920934#step:5:816
> 
> > test.inx : error 1199: The syntax 'DIRID 13 (CopyFiles)' was introduced in OS version 10.0.16299, but DDInstall sections utilizing the syntax will install on earlier OS versions. Those DDInstall sections should be restricted to only install on 10.0.16299 or higher using a TargetOSVersion decoration.
> 
> Could you advise?

I hope this helps: https://github.com/microsoft/Windows-driver-samples/issues/843

huangqinjin:
> I hope this helps: [microsoft/Windows-driver-samples#843](https://github.com/microsoft/Windows-driver-samples/issues/843)

Thanks, it works.

Dockerfile.clang
```diff
@@ -20,6 +20,7 @@ COPY test/hello.c ./
 
 RUN \
     for arch in x86 x64 arm arm64; do \
+        ([ $arch = arm ] && printf "%s\n" 10.0.26100.0 $(bash -c ". /opt/msvc/bin/$arch/msvcenv.sh && echo $SDKVER") | sort -VC) || \
```
mstorsjo, line 23: These invocations here need escaping the `$` in `$SDKVER`, otherwise `$SDKVER` gets expanded before invoking `bash`.

Overall - this expression for comparing version numbers does work, but is quite complex (including the nuances for getting the variable escaping correct); I wonder if it'd be simpler to do something like `[ -d /opt/msvc/kits/10/lib/*/um/arm ]` instead?

vsdownload.py
```diff
@@ -835,6 +848,12 @@ def moveVCSDK(unpack, dest):
         if args.with_wdk_installers is not None:
             unpackWin10WDK(args.with_wdk_installers, unpack)
 
+        if sys.platform != "win32":
+            # Wine doesn't support dependentAssembly yet.
```
mstorsjo, line 855: Can you add some comments (in the commit message?) about what has changed that causes this to be required now, when it wasn't required before? Is it some change in the newer versions of the WDK stuff? Something else?

vsdownload.py
```diff
@@ -723,6 +724,18 @@ def extractPackages(selected, cache, dest):
...
+    if not os.path.isfile(app + ".config"):
+        return
+    dest = os.path.dirname(app)
+    ns = "{urn:schemas-microsoft-com:asm.v1}"
+    configuration = ET.parse(app + ".config")
+    for codeBase in configuration.findall(f"./runtime/{ns}assemblyBinding/{ns}dependentAssembly/{ns}codeBase/[@href]"):
+        href = codeBase.attrib["href"].replace("\\", "/")
+        src = os.path.join(dest, href)
+        if os.path.isfile(src):
+            shutil.copy(src, dest)
```
mstorsjo, line 740: Can you add some comments/examples about which files end up copied, from where, in this case?

vsdownload.py
huangqinjin: See https://github.com/mstorsjo/msvc-wine/actions/runs/15065330342/job/42348765451#step:5:584.

I have updated commit message.

> vsdownload: Copy dependent assemblies to MSBuild.exe directory
> 
> MSBuild.exe depends on OpenTelemetry assemblies since VS 17.14.
> Their locations are specified in the MSBuild.exe.config file:
>   \<dependentAssembly>
>     \<assemblyIdentity name="OpenTelemetry.Api" publicKeyToken="7bd6737fe5b67e3c" culture="neutral" />
>     \<codeBase version="1.0.0.0" href="..\OpenTelemetry.Api.dll" />
>   \</dependentAssembly>
> 
> Unlike other dependent assemblies, such as Microsoft.Build.dll,
> which are duplicated in Bin/amd64 and Bin/arm64 directories,
> OpenTelemetry assemblies are found only in the Bin directory.
> 
> Since Wine does not yet support dependentAssembly,
> dependencies must be manually copied to the app directory.

Hopefully it is clear enough.

vsdownload.py
huangqinjin: I updated commit message to include the example OpenTelemetry.Api.dll.

Dockerfile.clang
huangqinjin: Yes. For this helloworld exmaple, a simpler check is enough, I changed to
```shell
[ ! -d /opt/msvc/kits/10/lib/*/um/$arch ]
```

But I would like to retain the current form in `test.sh`, which explicitly excludes unsupported tests to help identify errors, such as failures in the installation script.

mstorsjo: Thanks, that sounds reasonable!

mstorsjo (approved):
LGTM, thanks!

PR #193, opened 2025-06-04
vsdownload: Only download the right SDK for MSVC 16/17 with --msvc-version

mstorsjo:
Previously, we'd end up downloading both the default SDK from the manifest, and the SDK that was default for that version of MSVC.

Then install.sh would pick the latest SDK out of the ones installed, which wouldn't be the right choice when we explicitly wanted an older one.

If selecting an implicit default SDK via --msvc-version, set this as the option as if set via --sdk-version, which forces other SDKs to be ignored.


huangqinjin:
The change looks good to me. 

But I cannot see `--msvc-version` would include the default SDK from the manifest. From the dep tree, the default SDK is the dependency of `Microsoft.VisualStudio.Workload.VCTools` and `Microsoft.VisualStudio.Component.VC.Tools.ARM64`, which are not included when `--msvc-version` is specified.

PR #194, opened 2025-06-17
Make MSBuild work out of the box 

huangqinjin:
MSBuild now works without any environment presets or extra command line arguments.
The props importing hierarchy plus key properties setting of a vcxproj is shown below.

```
vcxproj
 <WindowsTargetPlatformVersion>
 Microsoft.Cpp.Default.props
  ImportBefore\Default\*.props
 Microsoft.Cpp.props
  Microsoft.Cpp.ToolsetLocation.props
   <PreferredToolArchitecture>
   Platforms\x64\Platform.props
    Microsoft.Cpp.Platform.props
     Platforms\x64\ImportBefore\*.props
     Platforms\x64\PlatformToolsets\v143\Toolset.props
      Platforms\x64\PlatformToolsets\v143\ImportBefore\*.props
      Microsoft.Cpp.MSVC.Toolset.x64.props
       Microsoft.Cpp.MSVC.Toolset.Common.props
        Microsoft.Cpp.WindowsSDK.props
         <WindowsSdkDir_10>
         <UniversalCRTSdkDir_10>
         <_LatestWindowsTargetPlatformVersion>
         <WindowsTargetPlatformVersion>
         $(UniversalCRTSdkDir)\DesignTime\CommonConfiguration\Neutral\ucrt.props
          <UCRTContentRoot>
          <TargetUniversalCRTVersion>
```

### DIFF
--- a/patches/MSBuild/Microsoft/VC/v170/ImportBefore/Default/msvc-wine.props
+++ b/patches/MSBuild/Microsoft/VC/v170/ImportBefore/Default/msvc-wine.props
@@ -1,0 +1,29 @@
+<!--
+ Copyright (c) 2024 Huang Qinjin
+
+ Permission to use, copy, modify, and/or distribute this software for any
+ purpose with or without fee is hereby granted, provided that the above
+ copyright notice and this permission notice appear in all copies.
+
+ THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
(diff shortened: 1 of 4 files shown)


mstorsjo (commented):
This looks good, but it would be nice to clarify in the PR description, that this is about MSBuild on Windows - it is already working in the wine wrapped setups on unix? (It is slightly clearer in the individual commit title, where it is prefixed with `patches:`, but perhaps it would be nice to clarify it further there as well.)

huangqinjin:
> This looks good, but it would be nice to clarify in the PR description, that this is about MSBuild on Windows - it is already working in the wine wrapped setups on unix? (It is slightly clearer in the individual commit title, where it is prefixed with `patches:`, but perhaps it would be nice to clarify it further there as well.)

Yes. It it for MSBuild on Windows, and also for directly running MSBuild in a Wine environment.

PR #195, opened 2025-06-22
github: Test targeting host only on ARM64 hosts

huangqinjin:
Certain packages required for the ARM64 toolchain may currently be brought in indirectly via dependencies of the VC workload or the x86/x64 toolchain. 

Add host-only tests on ARM64 hosts to identify any missing packages, to help us eventually eliminate the installation of the VC workload and the x86/x64 toolchain.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,6 +62,32 @@ jobs:
       # Intentionally not storing any artifacts with the downloaded tools;
       # the installed files aren't redistributable!
 
+  test-msvc-wine-linux-arm-target-host-only:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -146,18 +172,24 @@ jobs:
     runs-on: windows-11-arm
     steps:
       - uses: actions/checkout@v4
-      # `platform.machine()` returns AMD64 instead of ARM64 for the pre-installed Python 3.9.
```
huangqinjin, line 149: Native ARM64 python 3.12 is now pre-installed.

PR #201, opened 2025-09-09
vsdownload: Add an option --insiders

mstorsjo:
This allows downloading the "insiders" version of VS 2026 (--major 18), analogue to the earlier preview versions.

Note that currently if installing this version, it ends up downloading both version 14.44.35207 and 14.50.35503, and `install.sh` picks the lower version - so to properly use it, one currently has to remove the older version after downloading, before running `install.sh`.

CC @huangqinjin


huangqinjin:
Why not simply map `--preview` to insiders for `major >= 18`.

PR #211, opened 2026-02-21
Bump major to 18

huangqinjin:


mstorsjo (approved):
LGTM, but a comment in the commit message about the check for cl.exe would be appreciated.

PR #214, opened 2026-03-23
Include the package version in the listing of --print-selection

mstorsjo:
This allows quickly inspecting e.g. the current version of the preview compiler, without actually downloading and unpacking it.


mstorsjo:
CC @huangqinjin

huangqinjin:
LGTM!

PR #216, opened 2026-03-31
Reapply "Revert "github: Install Wine Mono 9.0 for macOS""

huangqinjin:
This reverts commit a40dd807ceaa189783766c447ea2c67b0d68f588.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,8 +96,6 @@ jobs:
           brew install wine-stable msitools ninja meson
           WINE=$(command -v wine64 || command -v wine || false)
           $WINE wineboot
-          curl -s -L -O https://github.com/madewokherd/wine-mono/releases/download/wine-mono-9.0.0/wine-mono-9.0.0-x86.msi
-          $WINE msiexec /i wine-mono-9.0.0-x86.msi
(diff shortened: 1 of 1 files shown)


mstorsjo:
The level of reapply/revert here is getting hard to track. What changed, why do we want to remove this again now?

huangqinjin:
Ok. Updated commit message to remove the `Reapply` level. macOS runners now install Wine 11.1 with Wine Mono 10.4.1. MSBuild works well without explicitly installing Wine Mono.

mstorsjo:
> Ok. Updated commit message to remove the `Reapply` level. macOS runners now install Wine 11.1 with Wine Mono 10.4.1. MSBuild works well without explicitly installing Wine Mono.

Thanks! Can you also include this info in the commit message? Then this would be good to merge.

huangqinjin:
> > Ok. Updated commit message to remove the `Reapply` level. macOS runners now install Wine 11.1 with Wine Mono 10.4.1. MSBuild works well without explicitly installing Wine Mono.
> 
> Thanks! Can you also include this info in the commit message? Then this would be good to merge.

Amended. A small correction: The wine-stable version is `11.0_1`.

PR #218, opened 2026-04-01
Install CMake (Windows version)

### LINKED ISSUE #118: MS CMake support
Has anyone figure out how to get cmake support working?
I know there is cmake on the linux side, but installing the microsoft distributed cmake would allow using the generator for VS's msbuild, which is needed in certain situations.

I have installed the cmake.exe using `--include-optional`, and then it is located somewhere in my `/opt/msvc` location, but when I try to run `wine $some_long_path/cmake.exe` I get
```cmd
ShellExecuteEx failed: File not found.
```

huangqinjin:
Closes #118.

### DIFF
--- a/test/test-cmake.sh
+++ b/test/test-cmake.sh
@@ -45,5 +45,10 @@ DIFF ninja-rerun.out - <<EOF || cat ninja-rerun.err
 ninja: no work to do.
 EOF
 
+CMAKEDIR=$(. "${BIN}msvcenv.sh" && echo $CMAKEDIR)
+if [ -d "$CMAKEDIR" ]; then
+    EXEC "" ${BIN}cmake.exe -GNinja -S "$TESTS" -B win-ninja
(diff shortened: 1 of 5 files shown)


huangqinjin:
Wine cannot find native `vcruntime140.dll` even it is in `WINEPATH`:
https://github.com/mstorsjo/msvc-wine/actions/runs/23837949065/job/69486153481#step:6:111
> 01e8:err:module:import_dll Library vcruntime140.dll (which is needed by L"C:\\windows\\system32\\MSVCP140.dll") not found

After reset `WINEDLLOVERRIDES`, tests pass on x64 host, but still failed on arm64 host:
https://github.com/mstorsjo/msvc-wine/actions/runs/23843724556/job/69505641267#step:6:113
> wine: failed to load /opt/wine/lib/wine/aarch64-windows/ntdll.dll error c00000bb

@mstorsjo any idea?

PR #222, opened 2026-05-24
Use python-msi for MSI extraction with msiextract fallback

playday3008:
On Fedora 44 (msitools 0.106.58), `msiextract` fails on several Win11 SDK MSIs (Application Verifier, MsiVal2) with `gsf_input_name: assertion 'GSF_IS_INPUT (input)' failed`, aborting `vsdownload.py`.

This adds optional `python-msi` (`pip install python-msi`) as a pure-Python MSI extractor, falling back to `msiextract` when not installed.

### DIFF
--- a/vsdownload.py
+++ b/vsdownload.py
@@ -844,10 +844,11 @@ def unpackWin10SDK(src, payloads, dest):
             if sys.platform == "win32":
                 # The path to TARGETDIR need to be quoted in the case of spaces.
                 cmd = "msiexec /a \"%s\" /qn TARGETDIR=\"%s\"" % (srcfile, os.path.abspath(dest))
+                with open(os.path.join(dest, "WinSDK-" + getPayloadName(payload) + "-listing.txt"), "w") as log:
+                    subprocess.check_call(cmd, stdout=log)
             else:
-                cmd = ["msiextract", "-C", dest, srcfile]
-            with open(os.path.join(dest, "WinSDK-" + getPayloadName(payload) + "-listing.txt"), "w") as log:
-                subprocess.check_call(cmd, stdout=log)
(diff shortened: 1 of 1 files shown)


mstorsjo:
This sounds interesting, thanks!

> On Fedora 44 (msitools 0.106.58), `msiextract` fails on several Win11 SDK MSIs (Application Verifier, MsiVal2) with `gsf_input_name: assertion 'GSF_IS_INPUT (input)' failed`, aborting `vsdownload.py`.

If this is the case, it sounds like it would be an msitools regression. I tested with Ubuntu 26.04 (with the bundled Dockerfile), which has msitools 0.106 - and there it seems to run fine.

I attempted changing the Dockerfile to run with Fedora 44, but for some reason, the version of Wine bundled there just hangs. But if I skip the Wine parts and just try to run `vsdownload.py`, I do indeed hit this issue. I'll have a look around...

mstorsjo:
Another surprising observation: This seems to work fine with both Fedora 43 and Fedora 45. In Fedora 43, it has got msitools 0.106.31, but Fedora 44 and 45 both seem to have the same version 0.106.58, while it seems to work fine in 45 but fail in 44 - that's strange...

playday3008:
@mstorsjo [That's](https://src.fedoraproject.org/rpms/msitools) even stranger

image

wroyca:
@playday3008 any chance you may upload failing artifacts to https://gitlab.gnome.org/GNOME/msitools/-/merge_requests/82, that would be very helpful!

playday3008:
> any chance you may upload failing artifacts to [gitlab.gnome.org/GNOME/msitools/-/merge_requests/82](https://gitlab.gnome.org/GNOME/msitools/-/merge_requests/82), that would be very helpful!

@wroyca Strangely I can't reproduce that issue, either Microsoft silently replaced msi packages or I can't explain what is going on

PR #225, opened 2026-05-27
Default to Preview MSVC for insiders (preview) channel

huangqinjin:


Matrix3600:
Currently, the --preview option allows to select the insiders channel and download any version of a specified package (preview or not), by its name, without specifying --msvc-version. I think this PR breaks this. Or we need an additional --insiders option to select the insiders channel.

huangqinjin:
> Currently, the --preview option allows to select the insiders channel and download any version of a specified package (preview or not), by its name, without specifying --msvc-version. I think this PR breaks this. Or we need an additional --insiders option to select the insiders channel.

Select single package is still supported (after #162).

PR #226, opened 2026-05-27
vsdownload: Don't include VC workload by default

huangqinjin:
Before:
  Selected 633 packages, for a total download size of 2.8 GB, install size of 9.1 GB

After:
  Selected 173 packages, for a total download size of 1.6 GB, install size of 5.3 GB


huangqinjin:
The interest thing after checking manifest file is that, only the latest (default version) ARM64-only toolchain is missing dependency on the VCRedist version package, Preview version or older versions are not.

PR #232, opened 2026-08-18
macOS Apple Silicon support

SamuelMarks:
* Update `install.sh` and `test/test-vcvars.sh` to treat Apple Silicon (`arm64`) as host `x64`, as Wine on macOS requires running x86_64 architecture wrappers natively under Rosetta 2.
* Add `msiextract` early error-handling to `vsdownload.py` with explicit instructions for `brew install msitools`.
* Document Homebrew requirements (`wine-stable`, `msitools`, `meson`, `llvm`) in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,8 @@ To install, just run the following two commands:
     ./install.sh <dir>
 
 The unpacking requires recent versions of msitools (0.98) and libgcab
-(1.2); sufficiently new versions are available in e.g. Ubuntu 19.04.
+(1.2); sufficiently new versions are available in e.g. Ubuntu 19.04. On macOS,
+you can install the prerequisites via Homebrew: `brew install wine-stable msitools meson`.
 
 After installing the toolchain this way, there are 4 directories with tools,
 in `<dest>/bin/<arch>`, for all architectures out of `x86`,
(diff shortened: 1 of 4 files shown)


mstorsjo:
As these are a couple separate changes, can you split it into separate commits for each individual type of change?

As for the changes to the shell scripts - there's no functional change involved in there though? Admittedly, this is somewhat by lucky coincidence as things stand right now - with `$(uname -m)` returning `arm64` on Darwin, it doesn't get classified as `aarch64` right now, and keep the default value of `x64` instead. This already does work on Darwin, and we already have CI tests that cover it - see https://github.com/mstorsjo/msvc-wine/actions/runs/26364422681/job/77605626222.

That said, the changes probably are reasonable, but they need to be labelled as the right thing - they're not fixing something that wasn't working before, they're just making sure we do classify it as aarch64, but then also decide to not use aarch64 binaries after all.

When Wine does end up working in aarch64 mode (which is coming at some point, see https://www.codeweavers.com/blog/mjohnson/2026/7/31/crossover-preview-the-right-to-bear-arm64-on-mac and https://list.winehq.org/hyperkitty/list/wine-devel@list.winehq.org/thread/CKG5CEN2BE5VRXZ7O7NX4YUSBH3247WH/) we probably do need a change like this, so this is probably a reasonable step towards that.

dplewis:
Can't install wine through homebrew new gatekeeping rules. May take a while to get these resolved
```
Error: wine-stable: Cask 'wine-stable' has been disabled because it does not pass the macOS Gatekeeper check! It was disabled on 2026-09-01.
```

SamuelMarks:
@dplewis Yeah I just ran into that same issue; downloaded from https://github.com/Gcenx/macOS_Wine_builds/releases and then `sudo xattr -dr com.apple.quarantine /Applications/Wine\ Devel.app`

